PR #1, opened 2015-05-06
Update ch6.md : forgotted coloring to js code.

syzer:


syzer:
also HUGE +1 for this book :)

DrBoolean:
Thanks!

syzer:
No.. thank you.
Do you think would be nice to show ppl how to use react.js, or any more common framework to make a website?

DrBoolean:
It is definitely a good idea. I hope to end each part with an example application. React is a great candidate for the next one.

PR #3, opened 2015-05-08
mention ensuring immutability via Object.freeze()

RebootJeff:


DrBoolean:
I was looking at it as `Object.freeze(2)`, which doesn't work. Good idea!

PR #4, opened 2015-05-08
clarify reference to earlier section

RebootJeff:
By the time I got around to the following line:

>So what is this undoubtedly nefarious *side effect* mentioned in the definition?

I was confused because I didn't know what "the definition" referred to. I was too busy thinking about mutability to recall that the top of the chapter starts with "the definition".
To be fair, I read at a lethargic pace, so the "issue" could be isolated ...to just me. :laughing:


DrBoolean:
I appreciate the feedback

PR #7, opened 2015-05-09
Copy tweaks for Chapter 2

mdavidgreen:
A few proposed rephrasings to enhance clarity


DrBoolean:
Thank you!

The [^] stuff is supposed to be for footnotes and shows up beautifully when i render it with marked2. Github on the other hand...

PR #9, opened 2015-05-10
More consinstent style with naming, missing ;

syzer:


DrBoolean:
One thing I hadn't mentioned was that I'm sticking with camelCase for functions, but snake_case for variables. The idea is that it's hard to tell when using first class functions everywhere, which is a function and which is a value.

syzer:
so kinda like python mixed style :)

gyzerok:
@DrBoolean I think that this could be achieved with using Flow. I found for myself that the most value from types are documentation feature and not static analysis :)

PR #11, opened 2015-05-10
fixed require modules out of scope

syzer:
require('../../support') 
also with new lodash code:

line 47:
```
var map = curry(function(f, xs) {
  return xs.map(f);
})
```
```var map = _.curry(_.map)```
there is also helper _.curryRight

### DIFF
--- a/ch4.md
+++ b/ch4.md
@@ -126,7 +126,6 @@ Let's acquire another essential tool called `compose`.
 ## Exercises
 
 ```js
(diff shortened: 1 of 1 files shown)


DrBoolean:
Ahh, just pasted in the exercise code directly. Best to remove the details like that. Thanks!

I think `_.map` is already curried in ramda so no need for the extra def.

PR #16, opened 2015-05-13
fix code typos in exercise1 of ch5

RebootJeff:
- rename variable (was a misnomer)
- add missing semicolon


DrBoolean:
Much nicer, thanks!

PR #25, opened 2015-05-17
Update ch6.md

mdavidgreen:
Normalize semicolons and variable assignments to avoid linting errors, and fix a little grammar


DrBoolean:
The semi-colons have bloomed.Thanks for all the work!

PR #26, opened 2015-06-09
Fixed the sq function.

hemanth:
It was just a relation and not a function before, fixed it with the appropriate images and code.

//cc @DrBoolean


DrBoolean:
Hi hemanth. I think there is some confusion.

The second example was to show a relation that wasn't a function for contrast. So you were, indeed correct - that is not a function, but on purpose.

The "doubles" relation was more about x2 than squaring since the sq root is only a function if it's domain is non-negative numbers: http://www.mathsisfun.com/sets/function-square-root.html

I appreciate the help. Is there something i can move to make it less confusing?

hemanth:
[L99](https://github.com/DrBoolean/mostly-adequate-guide/blame/3584968463b308c3456e34093168bba9021c0104/ch3.md#L99) was the confusing part.

> [^It appears this function doubles it's input].

But, there is no image of the function describing double, even though there was a carrot indication.

DrBoolean:
Ahh, I see. I'll try to reword that or make the example less similar to the graphic. I dig the updated graphic you did, we'd just have to make it double instead of square

hemanth:
:+1:

PR #27, opened 2015-06-09
Fix memoize example

pgilad:
`String(arguments)` will return `[object Arguments]` which in turn will return the same result for every argument.

Using `JSON.stringify` is an easy alternative solution.

### DIFF
--- a/ch3.md
+++ b/ch3.md
@@ -153,7 +153,7 @@ var memoize = function(f) {
   var cache = {};
 
   return function() {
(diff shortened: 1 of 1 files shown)


pgilad:
Might collide with: https://github.com/DrBoolean/mostly-adequate-guide/pull/9

DrBoolean:
Thanks!

pgilad:
:smile_cat:

PR #34, opened 2015-07-14
Fix typo.

schmich:
Fix typo.


DrBoolean:
Oh man, that's an embarassing one :)

PR #43, opened 2015-07-15
Match type in text with type in example code

cust0dian:
Example `getTwenty` function returns `Maybe(String)`, not `Maybe(Account)`


DrBoolean:
Thank you!

PR #46, opened 2015-07-15
Correcting the result

kdamball:
Because the object has no `age` property, `.map(_.prop('age'))` will return `Maybe(undefined)`.

### DIFF
--- a/ch8.md
+++ b/ch8.md
@@ -115,7 +115,7 @@ Maybe.of(null).map(match(/a/ig));
 //=> Maybe(null)
 
 Maybe.of({name: "Boris"}).map(_.prop("age")).map(add(10));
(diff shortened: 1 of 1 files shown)


DrBoolean:
Thanks for the help! I have to look out for `Maybe(null)` vs `Maybe(undefined)` throughout now...

kdamball:
What about returning `Maybe.of(null)` for `.map` for when `Maybe.isNothing() === true`?

```javascript
Maybe.prototype.map = function(f) {
  return this.isNothing() ? Maybe.of(null) : Maybe.of(f(this.__value));
}
```

This way, all the code is correct.

DrBoolean:
That's a good call. More explicit too.

PR #47, opened 2015-07-15
Not enough onion peeling

kdamball:


DrBoolean:
This is actually supposed to only join 1 layer. I threw that in to demonstrate. Maybe I should put a comment in like
`Task(Task("sewers")) <--- one layer

kdamball:
I had a feeling this function was different but I wasn't sure. Why would `Task(Task("sewers")) !== Task("sewers")`? Merely curious about this onion.

PR #57, opened 2015-07-17
Update log :: String -> IO String

christiantakle:
`log :: String -> IO String` to `log :: String -> IO ()`
`console.log` returns `undefined ` which I think is better represented by `()`

Which mean that log should be moved furthest to the left in the composed function or change the implementation of `log` to return the string e.g.

```js
// log :: String -> IO String`
 var log = function(s) {
   return new IO(function() { console.log(s); return s; });
 }
```
I will update the PR when I know how to progress.

### DIFF
--- a/ch9.md
+++ b/ch9.md
@@ -125,7 +125,7 @@ IO.prototype.join = function() {
 Again, we simply remove one layer. Mind you, we have not thrown out purity, but merely removed one layer of excess shrink wrap.
 
 ```js
-//  log :: String -> IO String
+//  log :: String -> IO ()
 var log = function(s) {
   return new IO(function() { return console.log(s); });
 }


DrBoolean:
Hell yes

christiantakle:
@DrBoolean 

I tried to make the types align but feel it obfuscates what you wish to teach and its still not safe-code, which we're selling here. `getItem` can have a malformed value for an existing key. `CSSProps` is just an alias for `Object`.

Maybe the solution is to sprinkling in some magic helpers and emphasise on intuition in the text.

```js
// tap :: (a -> IO ()) -> a -> a
var tap = curry(function(f, it) { f(it); return it; });

//  log :: a -> IO (a)
var log = function(x) {
  return new IO(function() { console.log(x); return x; });
}

//  setStyle :: CSSProps -> DOM -> String -> IO ()
var setStyle = curry(function(props, dom, s) { dom.style[s] = props[s]; });

//  setStyles :: Selector -> CSSProps -> IO DOM
var setStyles = curry(function(sel, props) {
  return new IO(function() {
    return tap(function(dom) {
      Object
        .keys(props)
        .map(setStyle(props, dom));
    }, $(el));
  });
});

//  getItem :: String -> IO String
var getItem = function(key) {
  return new IO(function() { return maybe('{}', id, Maybe.of(localStorage.getItem(key)));
  });
}

//  applyPreferences :: String -> IO DOM
var applyPreferences = compose(join, map(setStyle('#main')), join, map(log), map(JSON.parse), getItem);


applyPreferences('preferences').unsafePerformIO();
// "{backgroundColor: 'green'}"
// <div style="background-color: 'green'"/>
```

PR #58, opened 2015-07-17
Chapter 3 it's / its

ultranaut:
Fixed several cases of "it's" that should "its"


DrBoolean:
Thanks a million - I always get that wrong!

PR #61, opened 2015-07-17
Fix more it's / its issues

ultranaut:
I think that's all of them.


DrBoolean:
Yey! Its wonderful. (j/k)

ultranaut:
Word.

PR #65, opened 2015-07-18
Fix onHoliday return type

henrist:


DrBoolean:
Thanks! lousy comments...

PR #66, opened 2015-07-18
Align types in join example

christiantakle:
This should be runnable (havn't tried). 

CSSProps to Object, readers might not be aware of type aliases at this point.

Added tap to make a runnable `setStyles` example. It might be better to remove it and use an `interim` variable if its confusing for readers.

### DIFF
--- a/ch9.md
+++ b/ch9.md
@@ -127,7 +127,7 @@ Again, we simply remove one layer. Mind you, we have not thrown out purity, but 
 ```js
 //  log :: a -> IO a
 var log = function(x) {
-  return new IO(function() { console.log(s); return x; });
+  return new IO(function() { console.log(x); return x; });
 }
 
 //  setStyle :: Selector -> CSSProps -> IO DOM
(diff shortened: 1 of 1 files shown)


DrBoolean:
Thanks so much for this!

I like your example a lot, but I do feel like it's harder to understand. The `log` function certainly breaks the flow which is the root of the issue (I think). So perhaps we just update `log` for now? I don't mind offloading setStyles to `$` and ignoring the return value for simplicity's sake, but I do love `tap` (fond memories of ruby).

christiantakle:
Okay so just change `log` to return `IO Object` and keep everything else as before?

I'm confused about the `$(selector).style` if its returning regular DOM `node.style(object)` will break – are you thinking of `jQuery.css` http://api.jquery.com/css/#css-properties

DrBoolean:
Ahhh, yeah! `css` is what I wanted for sure.

Yeah, i think that and `log` returning its value and the signature being: `a -> IO a` would be a great fix. More strictly we could say `Showable a => a -> IO a`, but js will log whatever so the constraint isn't too necessary.

christiantakle:
I agree - I was thinking about making `log`s type more generic but I refrained since I thought it might be easier to follow the explicit types more rigidly.

I also changed `$`to `jQuery` just so its clear where things are comming from. There was a homemade version of `$` earlier in the guide. If you prefer `$` let me know.

DrBoolean:
This is so great! Thanks a bundle

PR #70, opened 2015-07-19
Fix minor typo: `map` to `chain`

christiantakle:


DrBoolean:
Hi @christiantakle. Thanks for all your help over the weekend (I still need to respond to the other issue).

In this case `map` will actually bundle the value back in an `IO` for us so it's what we want. The example is trying to demonstrate we can either `chain/of` of `map`

christiantakle:
You're right - I was overzealous and didn't see that its explicit in the text above the example. Somehow I only looked below it :)

PR #73, opened 2015-07-19
Minor style edits, mostly paragraph breaks

TehShrike:
A couple extra spaces, but mostly just newlines to create paragraph breaks.

If slim readable paragraphs aren't your style, let me know - otherwise I'll probably be making more pull requests like this one against other chapters.

### DIFF
--- a/ch1.md
+++ b/ch1.md
@@ -2,9 +2,15 @@
 
 ## Introductions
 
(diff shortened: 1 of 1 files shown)


DrBoolean:
Much nicer on the eyes

PR #75, opened 2015-07-19
Added '_.' to make exercises in run in node.js

jazmit:
Alternative would be to define

    match = _.match
    filter = _.filter
    reduce = _.reduce
    split = _.split


DrBoolean:
Thanks!

PR #78, opened 2015-07-20
Fixed confusing concat method

gyzerok:
Fixed confusing concat method


DrBoolean:
Oops, didn't see this one! Thanks for that. I need to go through and add `_` pretty much everywhere.

PR #85, opened 2015-07-21
Instructions to setup immediate feedback loop while answering questions

manuganji:
For a way to stay motivated throughout the exercises.


DrBoolean:
This is awesome. Thanks! Can we make it work with just `npm install` from the part1_exercises/package.json?

manuganji:
I'll try and make it work with package.json. Glad that you liked it! :)

manuganji:
Hey @DrBoolean how does it look now?

PR #86, opened 2015-07-21
Fix typo in ch8.md

SyntaxBlitz:
Original: As it happens, `map` is fully equip to delegate...

Revision: As it happens, `map` is fully equipped to delegate


DrBoolean:
Thanks!

PR #87, opened 2015-07-21
Fix typo in ch5

MartijnDwars:
Original:  Since it doesn't matter how we group our calls to `compose`, the result will be the same. That allows **use** to write a variadic compose and use it as follows:

Refactored: Since it doesn't matter how we group our calls to `compose`, the result will be the same. That allows **us** to write a variadic compose and use it as follows:


DrBoolean:
Thanks!

PR #90, opened 2015-07-21
Added _. to chapter 5 so exercises run in Node.js

jazmit:


DrBoolean:
Did they not work before? I worry because the later examples are dependent upon `support`

jazmit:
This is only for when copying and pasting the exercise code into the node.js repl, which some people might do just to dip into the exercises. (I did this before I realized that there were specs hiding in the `/code` folder!).  The stand-alone exercise file works just fine.

jazmit:
Really looking forward to chapter 10!

jazmit:
Hmm, the `support` requirement means the later exercises won't work copy and paste, perhaps there should be some brief instructions in the book for setting up the exercises?

PR #92, opened 2015-07-22
add missing semicolons

chicoxyzzy:


jirutka:
This just adds unnecessary and useless noise. Semicolons are *not* required nor needed in JS, even Standard (code-style) [recommends to not use them]( https://github.com/feross/standard/blob/master/RULES.md#automatic-semicolon-insertion-asi).

chicoxyzzy:
There are much more style guides which recommend semicolon. Anyway consistency is the reason.

DrBoolean:
Thanks! Yeah I started without semi-colons and it was pointed out to me that most people's linters get all worked up without them. A this point we already made the effort to add them so let's press on.

PR #93, opened 2015-07-22
Chapter 2 paragraph breaks and other minor changes

TehShrike:
Mostly just whitespace and capitalization.

I did change "something more novice" to "something more fiddly" at the end there because the former sounds unnecessarily disparaging.


DrBoolean:
Nah, i like it!

PR #94, opened 2015-07-22
Chapter 3 paragraph breaks, small changes

TehShrike:
Including...

- paragraph breaks!
- added some whitespace that is, if I remember correctly, significant in Standard Markdown parsers
- removed unnecessary indentation in code blocks
- removed redundant-seeming "to say the least"


DrBoolean:
Dig it. Thanks!

PR #97, opened 2015-07-22
Replace occurences of "new IO" with "IO.of"

mallowigi:
Since you specified that the "of" function was for this purpose, you probably forgot to use it there.

### DIFF
--- a/ch8.md
+++ b/ch8.md
@@ -375,7 +375,7 @@ Let's see it in use:
 
 ```js
 //  io_window_ :: IO Window
-var io_window = new IO(function(){ return window; });
+var io_window = IO.of(function(){ return window; });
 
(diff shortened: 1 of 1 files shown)


DrBoolean:
Ahh, perhaps, I need to clarify in the text. The `of` here will lift the entire function in as its value. So we'd end up with a function that returned window as the output of `.unsafePerformIO()`.

Normally, we could write `IO.of(window)` directly, but then `window` would be accessed prematurely - hence the need for the function around it.

So: `IO.of(window) != new IO(function(){ return window })` because the one on the right is lazy and `IO.of(function(){ return window })` lifts the entire function as it's value.

Thanks for bringing this up - I need to add some clarifying info in the Pointed section.

PR #99, opened 2015-07-22
Remove unnecessary semicolons at EOL in JS examples

jirutka:
Semicolons are really not necessary in well-formated JS, it’s just a syntactic noise. You can read a comprehensive explanation in [Standard](https://github.com/feross/standard) (code-style) [here](https://github.com/feross/standard/blob/master/RULES.md#automatic-semicolon-insertion-asi).

This PR is a direct reaction to #92. Now you can simply choose if you want to consistently add, or remove unnecessary semicolons.

Don’t hesitate to just close this PR without any reaction if you don’t want to remove semicolons, I’m perfectly fine with that! It’s naturally up to you how you choose to format code in your book. However, @chicoxyzzy is right that you should be consistent, if possible.

I’m looking forward to the Part 3 of your wonderful book! :smile_cat:

### DIFF
--- a/ch1.md
+++ b/ch1.md
@@ -20,24 +20,24 @@ Let's start with a touch of insanity. Here is a seagull application. When flocks
 
 ```js
 var Flock = function(n) {
-  this.seagulls = n;
-};
+  this.seagulls = n
+}
 
 Flock.prototype.conjoin = function(other) {
-  this.seagulls += other.seagulls;
-  return this;
-};
+  this.seagulls += other.seagulls
+  return this
+}
 
 Flock.prototype.breed = function(other) {
-  this.seagulls = this.seagulls * other.seagulls;
-  return this;
-};
+  this.seagulls = this.seagulls * other.seagulls
+  return this
+}
 
-var flock_a = new Flock(4);
-var flock_b = new Flock(2);
-var flock_c = new Flock(0);
+var flock_a = new Flock(4)
+var flock_b = new Flock(2)
+var flock_c = new Flock(0)
 
-var result = flock_a.conjoin(flock_c).breed(flock_b).conjoin(flock_a.breed(flock_b)).seagulls;
+var result = flock_a.conjoin(flock_c).breed(flock_b).conjoin(flock_a.breed(flock_b)).seagulls
 //=> 32
 ```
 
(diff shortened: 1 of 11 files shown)


DrBoolean:
Hi @jirutka! I fully agree. Someone a few weeks back went and added every semicolon throughout since the es-linters have taken over the world apparently. I appreciate the sentiment and work on this and fully agree with it, but I think I lean toward biting the bullet as something like pasting into jsbin will start spewing red immediately without them.

gyzerok:
I consider this a bad codestyle. Check linters configuration of most huge companies. Airbnb for example.
Especially this is not a codebase, but book. We shouldn't confuse readers.

gyzerok:
And semicolons at the start of the line is the most terrible thing i've ever seen.

sadiqevani:
Yo,

You do understand that those semicolons are not mandatory because the compiler add them automatically by itself. But this doesn't mean you should write those semicolons.

One of the major issues when not adding semicolons is when you return an object:

let say you write something like:

```
function returnObject() {
  return 
{
  name: sadi;
}
}
```

When compiled the following functions will be generated like this:

```
var returnObject = undefined;
returnObject = function () {
  return; // Note the semicolon here.
{
name: sadi;
};
};
function returnObject() {
  return 
{
  name: sadi;
}
}
```

kwijibo:
This seems the same core decision as the issue someone else raised wanting the examples rewritten as ES6.
Should code be written in most cutting-edge best style? Or written to be most widely understood by devs and tools?
It's an interesting question, but to an extent, this is bikeshedding. The important thing is the explanation of the concepts.

gyzerok:
@kwijibo you are trying to compare different things. ES6 is current standard of the language. [JavaScript Standard Style](https://github.com/feross/standard) is just another subjective code-style and nothing more.

alex-wilmer:
Semi-colons as a debate is definitely bike-shedding. Maybe we could make a "style-type" or "ES version" toggle button. I'd be into coding that.

Btw, @DrBoolean keep up the excellent work!

DrBoolean:
Thanks! That'd be really slick, but probably unnecessary. I think ES version might be more than just a toggle since we'd want to take advantage of some extra features and the semi colons are just to appease the masses. I appreciate the thought though!

gyzerok:
@sadiqevani +1

PR #104, opened 2015-07-22
Filled in missing bits for monads exercise 4

jazmit:
I'm not sure if you wanted to add type signatures to all three functions here, to give readers a chance to appreciate the type 'dimensional analysis', or to make it harder by forcing them to figure out the types themselves..

There seemed to be a missing line in the description, I had a stab at filling it in, let me know if that wasn't what you meant!


DrBoolean:
Dude, awesome. Thanks!

PR #110, opened 2015-07-24
Fix inspect.

tonyday:


DrBoolean:
Thank you! I never remember the order of those assertions.

PR #115, opened 2015-07-27
fix ch9 typos

akatopo:
* Since `IO` is graciously lending us ~~the~~ its value
* They come to the rescue **when** a value finds itself jailed


DrBoolean:
Thanks!

PR #117, opened 2015-07-27
Change flickr protocol to https

igorrKurr:
Change flickr protocol to https to be able to run demo on services like jsbin.

Also jsbin itself as a bonus :smile: https://jsbin.com/wezimeluzu/edit?html,js,output

P.S. Many thanks for this project, it is really great job! :+1:


DrBoolean:
Appreciate the help!

PR #127, opened 2015-08-03
improve `trace` example

### LINKED ISSUE #124: What's the regex supposed to do in dasherize in chapter 5?
```javascript
var dasherize = compose(join('-'), map(replace(/\s{2,}/ig, ' ')), split(' '))
```

I can't think of any useful application :confused:

despairblue:
fix #124

### DIFF
--- a/ch5.md
+++ b/ch5.md
@@ -159,25 +159,25 @@ var trace = curry(function(tag, x){
   return x;
 });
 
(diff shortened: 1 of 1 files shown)


despairblue:
@DrBoolean thoughts?

DrBoolean:
Thank you man!

PR #137, opened 2015-08-11
Fix incorrect anchor link of ch7

yang-wei:
tales-from-cryptic -> tales-from-the-cryptic


DrBoolean:
Ahh, i was wondering what that was! Thanks

PR #138, opened 2015-08-14
Just a minor change

GaurangTandon:
Moved the colon up one line. That separate colon all alone on one line confused me for a moment. I think it's better to have it right after `Maybe.of({balance: account.balance - amount})`


DrBoolean:
Thanks!

PR #151, opened 2015-08-31
change "map" in Bonus 1 to be Ramda's map

calesce:
For consistency's sake, as well as the fact the Array.prototype.map doesn't work. I think that's because it needs to be curried first, which Ramda does by default? (still learning!)


DrBoolean:
Thanks!

PR #158, opened 2015-09-05
Added Immutable js dependency and getters for accessing values from map

kelsadita:
Updated sample code of chapter 3 to include Immutable js dependency and used getter function to access values from `Immutable.Map`, since its only way to retrieve values from them.


DrBoolean:
Thanks! Totally overlooked the getters!

PR #172, opened 2015-09-14
Typos Fix

legomushroom:
misspelled “safely”
misspelled “implement”


zerkms:
Btw, you can make several commits - github will capture them all successfully. You can even `squash` them and it would be captured as well.

legomushroom:
@zerkms thanks. ugh found more in exercises files

PR #179, opened 2015-09-29
In curry exercise set initial max value to -Infinity instead of zero

bgryszko:
So example will also give correct result for this set for example. `[-10, -20, -30, -50] => -10`, now it gives `0`.


DrBoolean:
Thanks again!

PR #180, opened 2015-09-29
Explain bonus 2 exercise for chapter about currying

bgryszko:
I know it’s a detail but IMO it can be unclear for the reader. Feel free to accept it or not. Thank you one more time for an awesome job!


DrBoolean:
Thanks!

PR #184, opened 2015-10-01
Normalize if statements to the conventional format and fix #183

### LINKED ISSUE #183: Conditional statement formatting
`if` statements and the following parenthesis should be separated by a space because it's the conventional style and because the other way makes them look like functions, which they are not. In other words, this:

```if (foo) { …```

Instead of this:

```if(foo) { …```

slikts:
Fix #183


DrBoolean:
Thanks!

PR #188, opened 2015-10-10
Simplify visual perception of findParam

kapral18:
If there are more than 3 bracket pairs on one line in functional coding style, visual decoding becomes much more challenging than the code itself. This is just a suggestion.

### DIFF
--- a/ch8.md
+++ b/ch8.md
@@ -419,7 +419,16 @@ var params = compose(toPairs, last, split('?'));
 
 //  findParam :: String -> IO Maybe [String]
 var findParam = function(key) {
-  return map(compose(Maybe.of, filter(compose(eq(key), head)), params), url);
+  return map(
+    compose(
+      Maybe.of, 
+      filter(
+        compose(eq(key), head)
+      ), 
+      params
+    ), 
+    url
+  );
 };
 
 ////// Impure calling code: main.js ///////


guumaster:
IMHO, I'd prefer to read it as a pipe from right to left. no indentation needed, no multi-lines.

kapral18:
@guumaster It's not about sequence of reading, but about correct nesting, visually it's hard-to-guts to sort out when the filter ends, I mean brackets start to divide-in-two.

DrBoolean:
I believe if we're going to do this here, we should do it across the book for consistency.

kapral18:
If we stick to the rule of <=3 brackets in a row,
we will only have a few changes throughout the book.

tarekkazak:
I have to agree with @guumaster. I find it easier to comprehend as it is.  I'd rather sort out nesting then to read code top-down then down-up for the logical flow. If it got longer, I wouldn't be against line breaks but for me (as I concede this is entirely a question of preference) splitting parens adds to the noise.

kapral18:
@tarekkazak I agree that reading from right to left is in simple cases without nesting very natural, but in complex cases with multiple nesting levels all this bracket mess just adds so much noise, so instead of reading the flow you are actually figuring it out, what is after what, what is in which group, when this starts when this ends. Generally speaking a visual mess.

Verical, indentation based grouping is pretty much natural and is meant to deal with such use cases.

tarekkazak:
@kapral18, understably what I say could only ever apply to simple cases. But strictly speaking, if we've written something that complex I would be inclined to break it up into smaller (possibly reusable) pieces before trying to make it more legible.

kapral18:
@tarekkazak well this is double-edged, still, in one case you have one-liner of (()()(()()()()(((())()()(())() in other case you have _tempvar1, _tempvar2, _tempvar3, _tempvar4..., al this can be avoided by just visually nesting it into indented groups, after some threshhold, the style of indentation can be discussed, it doesn't matter, what matter is the amount of code to be focused on at a time, I am not a machine, I need to understand wtf is meant here. Although I am pro semantical aggregation of unit functionality into separate variable, I am forcibly against their overproduction and overuse.

tarekkazak:
@kapral18. First of all I would like to say this is a very enjoyable debate so far, design and structure are after all at the heart of happy coding. Secondly, my point was one of design in general. As in, if my code has gotten very complex in a specific area, I would seek to refactor it beyond the context of whatever function its in before worrying about visual style. So for me whether it's  (()()(())) or even tempvar1-4, we are identifying a greater problem.  One that should be resolved first and in my opinion, once resolved, the code should leave very few instances where visual style becomes an impediment.  But in the case that it is, then, to my mind, it is really a question of preference and habit.

 I will admit however that I shouldn't knock before I try it. :)

PR #190, opened 2015-10-11
Update ch8.md: findParam output incorrect

kapral18:
The result is Maybe of nested array, not just array. Here look: http://plnkr.co/edit/08qaFHro9onCy0kHL9TW.


DrBoolean:
Killin' it.

PR #196, opened 2015-10-20
Fix link to fpcomplete in ch8.md

fix-fix:
I don't know what it's supposed to look like (either link or footnote), but now it doesn't neither way.

This issue was introduced in 50b0d91f and I think this commit still doesn't make your text to contain proper footnotes.

The proper way to use footnotes in markdown and Gitbook is described in [Gitbook docs section about footnotes](https://help.gitbook.com/format/markdown.html#footnotes), so you might look at it and make changes accordingly.


DrBoolean:
Hey there!

Thanks for the fix - it really should be a link there.

I did a brute force regex fix and probably introduced a couple of these. I have to go through and check.

The footnotes were in the format mentioned by the gitbook docs, but epub had major issues so I removed them.

PR #197, opened 2015-10-21
Revert "Chapter 10 exercises fix"

fix-fix:
Reverts MostlyAdequate/mostly-adequate-guide#181

That change in 8e2189 broke the spec for Exercise 3 in ch10 and added inconsistency with other exercises.


DrBoolean:
Fair enough. It really should say tasks, but whatever if it passes

PR #203, opened 2015-11-07
ch04 - introductory sentence in lodash.curry ex

sdtsui:
Before: 
> Here we've made a function `add` that takes one argument and return a function.
After:
> Here we've made a function `add` that takes one argument and return*s* a function.


DrBoolean:
THanks!

PR #211, opened 2015-12-01
futility -> utility -- Typo?

christiantakle:
The use of this `futility` seems confusing to me. Is it the correct word?

Maybe it should have been `utility`

http://www.thesaurus.com/browse/futility


DrBoolean:
Yes indeed I meant "ineffectiveness"

PR #212, opened 2015-12-03
add missing Ramda reference

timuric:
are functions like `split` deliberately missing `_. `?


DrBoolean:
Thank you! I need to go and do this throughout the book. The original idea was that they were, in one way or another, previously defined via the book and not needed. But yeah, ramda makes way more sense.

Issue #83, opened 2015-07-21
Using ES6


gyzerok:
What you think about rewriting all the code examples in the book in ES6? It feels more functional than ES5.

gyzerok:
I can do it if you want.

sdwolf:
Better than rewriting why not provide the ES6 alternative alongside the ES5. I would be :+1: for that.

hemanth:
:+1: Re-write, there are few references at [1-liners](https://github.com/orgs/1-liners/)

jirutka:
+1

DrBoolean:
I would rather not put the two together. My preference would be a "second edition" fork for es6. That way people are not forced to use babel or bombarded with examples.

reflectiveSingleton:
+1 for the second edition...a lot of people have yet to learn the new ES6 syntax and features.

I think it is better not to necessarily require them to be super up to date on their language knowledge (as well as build process - babel/etc).

gyzerok:
@DrBoolean Is there any reason to support version for out of date standard? I think only lazy didnt use babel. This book is not for lazy people :)
But you anyway want to keep 2 additions let me say ES6 should be primary target and ES5 addition only for backward compatibility.

sdwolf:
@gyzerok How about front end javascript (browsers). It's gonna take a while for them to be ES6 compliant. 
While it would not be hard for me to translate to ES5 for my own code when I need to, I don't think this is applicable for everybody.

gyzerok:
@sdwolf A lot of software are still written with .NET 2.0. Does it mean we need to write new books for it? Out of date langs are used all over the place, not only in web world (javascript).

sdwolf:
@gyzerok just don't rush to take ES5 out. Having a reference to the "old way of doing things" can be useful in learning.

Silverwolf90:
Just to add this a bit... using `import` I think you can reduce the boilerplate of adding `_.` to everything.  I do this at home with `lodash-fp` as I find `_.` everywhere to add a lot of noise. Although I could see, that since this is a book, perhaps that might just add additional complexity to understand the examples.


```javascript
import { compose, prop, map} from 'ramda';

let mediaUrl = compose(prop('m'), prop('media'));
let srcs = compose(map(mediaUrl), prop('items'));
let images = compose(map(img), srcs);
let renderImages = compose(Impure.setHtml("body"), images);
let app = compose(Impure.getJSON(renderImages), url);
[...]

PR #234, opened 2016-01-06
Factor out `Maybe.of`

christiantakle:
Would I be crazy if I update alle the guides examples to use `=>` syntax? I feel it would make the examples more clear by removing noise.

If you wouldn't mind that change let me know if you prefer a particular style. I would properly make a PR for a chapter and then we could discuss it further.

```js
//  withdraw :: Number -> Account -> Maybe(Account)
var withdraw = curry((amount, {balance}) =>
  Maybe.of(balance >= amount? {balance: balance - amount} : null);

// or
var withdraw = // :: Number -> Account -> Maybe(Account)
  curry((amount, {balance}) =>
    Maybe.of(balance >= amount? {balance: balance - amount} : null);
```

### DIFF
--- a/ch8.md
+++ b/ch8.md
@@ -101,7 +101,7 @@ Maybe.prototype.isNothing = function() {
 }
 
 Maybe.prototype.map = function(f) {
-  return this.isNothing() ? Maybe.of(null) : Maybe.of(f(this.__value));
+  return Maybe.of(this.isNothing()? null : f(this.__value));
 }
 ```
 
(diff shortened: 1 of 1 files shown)


KtorZ:
Hey !
Your PR seems to be related to the issue #83. 
Nevertheless, we've no news about an existing fork with an ES6 version as discussed. If you want to start one, that would be great :)

christiantakle:
@KtorZ Cool, I'm starting here #235. Just need guidance on style then I will finish it this weekend.

PR #237, opened 2016-01-08
Fix typos in Chapter 8

secobarbital:


KtorZ:
Thanks. 
We'll merge #242 which cover the same typos and a bit more.

PR #243, opened 2016-01-22
Use this instead of referencing by name

timuric:
It might feel more "natural" to use `this` keyword in this instance. Since it would make refactoring easier and allow us to reuse `of` function for example.

```js
var of = function(x) { return new this(x); };

var Container = function(x) {
  this.__value = x;
}
Container.of = of;

var Jar = function(x) {
  this.__value = x;
}

Jar.of = of;
```

### DIFF
--- a/ch8.md
+++ b/ch8.md
@@ -13,7 +13,7 @@ var Container = function(x) {
   this.__value = x;
 }
 
-Container.of = function(x) { return new Container(x); };
+Container.of = function(x) { return new this(x); };
 ```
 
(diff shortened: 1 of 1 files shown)


KtorZ:
Not sure about this :|
Those code samples don't aim at being part of an architecture, or being reused. They just illustrate the words of the book. Keeping them simple and straightforward might be a better idea so that readers minds don't have to jump here and there to remember to what part of the code something is referring to.

There are already some complaints about libraries used like `ramda` and some functions defined in previous chapters and a bit lost in the flow.

timuric:
@KtorZ I see, thanks for the explanation!

PR #249, opened 2016-02-24
ch4: fix an installation snippet

eush77:
Since the earlier example in this chapter uses `lodash.curry` since #69, the instruction to `npm install lodash` should be updated, too.

### DIFF
--- a/ch4.md
+++ b/ch4.md
@@ -87,7 +87,7 @@ censored('Chocolate Rain');
 
 What's demonstrated here is the ability to "pre-load" a function with an argument or two in order to receive a new function that remembers those arguments.
 
-I encourage you to `npm install lodash`, copy the code above and have a go at it in the repl. You can also do this in a browser where lodash or ramda is available.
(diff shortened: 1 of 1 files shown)


KtorZ:
Hey!
Thanks for the proposal, however, because `lodash` is used everywhere in the book with different functions from the library, the user actually needs a complete install (or at least, an install of all required `lodash` functions which would be a hell to list and manage throughout the writing).

However, after having a look at #69, we definitely need to change the require to:

```js
require('lodash/curry')
```

The `lodash` library is really well built and can allow us to require only the needed functions (which is what we're currently interested in). We don't need to install another package, we can merely import the needed function and bundling tools like `browserify` of `webpack` will take advantage of that. 

Although the installation instructions won't change because they address a global concern, I (or feel free to PR :) ) will change the current `lodash` imports in favor of the above form.

eush77:
Actually, I haven't found any other uses of `lodash` in examples throughout the book. I submitted a PR for this particular case (#251).

PR #258, opened 2016-03-14
Fix type signature of renderPage

safareli:


DrBoolean:
I think the type signature was good as it stood on this one. Thanks for all your help, by the way

safareli:
Welcome

PR #280, opened 2016-06-14
Fix a typo

vitkarpov:
There's only one spade :)


nicolasfont:
That seems intentional, not a typo. See https://en.wikipedia.org/wiki/Call_a_spade_a_spade.

PR #281, opened 2016-06-14
make argument to reduce match the ch4.md value

radiofreejohn:
Is it worthwhile making this 1:1 with the text?

For example, in the text it shows:
```
// REFACTOR THIS ONE:
var max = function(xs) {
  return _.reduce(function(acc, x) {
    return _keepHighest(acc, x);
  }, -Infinity, xs);
};
```

versus the above without `_.reduce`


DrBoolean:
Thanks!

PR #282, opened 2016-06-14
Fix link to parametricity

radiofreejohn:
kept the italics formatting, but added correct link format, and changed to https since wikipedia redirects to https anyways


DrBoolean:
Thanks!

PR #273, opened 2016-05-02
Update to ES6 | ES2015 (revived)

christiantakle:
I've accidentally deleted my previous repository #235 
### From old PR
- [x] Chapter 1
- [x] Chapter 2
- [x] Chapter 3
- [x] Chapter 4
- [x] Chapter 5
- [x] Chapter 6
- [x] Chapter 7
- [ ] Chapter 8 (updated but needs feedback)
- [ ] Chapter 9 (updated but needs feedback)
- [ ] Chapter 10 (updated but needs feedback)
- [x] Write tests for all examples or make all examples runnable?
- [ ] Update FAQ
- [ ] Update images (app example)
- [ ] What to do with fn declared by `const` with the same name? index e.g. `f1, f1` them or use `let` instead

Links that need to point to the correct location in Master. Might be easier to fix after merge.
- [ ] Update paths to exercises
- [ ] grep for `part1_`
- [ ] grep for `part2_`

I think we should update the text and provide a better OOP example. The code is still very verbose. We could also keep the current mutating code and just call it imperative instead of OOP.

``` js
class Flock {
  constructor(n) {
    this.seagulls = n;
  }

  conjoin({seagulls}) {
    return new Flock(this.seagulls + seagulls);
  }

  breed({seagulls}) {
    return new Flock(this.seagulls * seagulls);
  }
}
```

Based on:  https://gist.github.com/eshacker/be153d852b9b995f42dd

PR #283, opened 2016-06-17
compose_exercises.js

loque:
Use ramda's toLower instead of toLowerCase (as it's not curryed) in Exercise 4.

### DIFF
--- a/code/part1_exercises/answers/compose/compose_exercises.js
+++ b/code/part1_exercises/answers/compose/compose_exercises.js
@@ -36,7 +36,7 @@ var averageDollarValue = _.compose(_average, _.map(_.prop('dollar_value')));
 
 var _underscore = replace(/\W+/g, '_'); //<-- leave this alone and use to sanitize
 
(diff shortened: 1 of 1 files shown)


loque:
I see now that these functions (toLowerCase, join, etc) are defined in `support.js`. I understand that now but it is confusing when copy-pasting the answers to compare with my own answers. My confusion is caused by the mixed use of ramda's functions and these custom functions (which already  have a ramda equivalent).

Wouldn't it be easier if only the ramda's functions where used?

christiantakle:
@loque I'm doing that in the base update of the book (https://github.com/MostlyAdequate/mostly-adequate-guide/pull/273/files#diff-f98d81b9fc29d69526247d3e2d20f70aR11). If you have any feedback please let me know in #273

KtorZ:
I'll merge it for now and work a bit on a rework of the exercises, 

I find it also a bit confusing t have a mix of ramda functions with support functions. We should go for either one of them I believe. However, the support functions are also referred to from the book. 
I'll open an issue to discuss the matter.

PR #291, opened 2016-07-20
Minor reformatting and added a Wikipedia link

rapPayne:
First attempt at helping. Minor edits. Moved part of long comments to the next line because the wrapped on narrower screen widths.  I also added a Wikipedia link for 'Cognitive Load' for those who don't know what that means.

### DIFF
--- a/ch2.md
+++ b/ch2.md
@@ -38,7 +38,7 @@ greeting('times');
 // "Hi times"
 ```
 
-In other words, `hi` is already a function that expects one argument, why place another function around it that simply calls `hi` with the same bloody argument? It doesn't make any damn sense. It's like donning your heaviest parka in the dead of July just to blast the air and demand an ice lolly.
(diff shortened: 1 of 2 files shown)


ch2.md
```diff
@@ -38,7 +38,7 @@ greeting('times');
 // "Hi times"
 ```
 
-In other words, `hi` is already a function that expects one argument, why place another function around it that simply calls `hi` with the same bloody argument? It doesn't make any damn sense. It's like donning your heaviest parka in the dead of July just to blast the air and demand an ice lolly.
+In other words, `hi` is already a function that expects one argument, why place another function around it that simply calls `hi` with the same bloody argument? It doesn't make any darned sense. It's like donning your heaviest parka in the dead of July just to blast the air and demand an ice lolly.
```
KtorZ: I think we lose a bit of the original prose here. The sense is indeed preserved yet I would rather keep the original form.

KtorZ:
Thanks for the contributions :)

PR #295, opened 2016-08-10
Modify functor_exercises.ex4 to avoid return undefined when input is 0.

lightPang:
Fix minor issue caused by insufficient condition.

Modify function functor_exercises.js under exercises and  answers to make function ex4 works perfectly as the same with our answers.

Previously, when input is zero, predefined function ex4 will return undefined when input is 0. After this commit, it will return 0 in the same case.

Sorry for deleting the "/" file as I'm using windows and will meet with fatal error when I try to remove it from commit.


code/part2_exercises/exercises/functors/functor_exercises.js
```diff
@@ -35,7 +35,7 @@ var ex3 = undefined;
 // Use Maybe to rewrite ex4 without an if statement
 
 var ex4 = function (n) {
-  if (n) { return parseInt(n); }
+  if (n !== null && n !== undefined) { return parseInt(n); }
```
safareli: you could do [!= null](http://stackoverflow.com/a/27757708/1798418)

lightPang: Thanks for your suggestion, but I think use a strict way to make up condition will be better, especially for the new souls.

FangfeiAsFey: Dear, 

Please take  a look at all  these things  I've just found, I think this is what  you were looking for. Check it out <http://best.mikefata.org/e7vl/6>

Thx, fey_y

tutukin: Shouldn't it be something like `if (typeof n !== 'number') {`?

KtorZ:
Closed in favor of #361.

PR #297, opened 2016-08-13
Replace non-existent toArray function

kilkelly:


KtorZ:
What about defining the function `toArray` since it's more readable and used at several places ?

```js
var toArray = Function.call.bind(Array.prototype.slice)

or 

function toArray (x) {
    return Array.prototype.slice.call(x)
}
```

?

fix-fix:
In ES2015 there is now an [`Array.from()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) function that behaves exactly like `toArray`.

KtorZ:
Yeah :/ 
But this version of the book sticks with ES5. 
It is a neat alternative for the ES6 / ES2015 version though (see #273)

safareli:
@KtorZ it's offtopic but i was not able to find any documentation of `Function.call/apply`nor in google nor in es6 spec. I know how `Function.prototype.call/apply` works, but, just `Function.call/apply` is kinda new and what's difference between them? or could you link an article/documentation about that topic. The only thing I found in es6 spec was  [26.1.1 Reflect.apply ( target, thisArgument, argumentsList )](http://www.ecma-international.org/ecma-262/6.0/#sec-reflect.apply)

KtorZ:
Same thing ^.^ ... This is just a shortcut. Since `Function` is in fact a JavaScript function, it does have the function prototype known as `Function.prototype`. Therefore, when you access `Function.call`, which doesn't exist a priori on the `Function` object, there's a lookup done in the prototype chain. The `call` method is defined in the function prototype so, no problem.

KtorZ:
Duplicate with #340

PR #302, opened 2016-08-27
Ch 8 - Fix type of ctmd.getCompose

gusvargas:
Not immediately sure what the intention of showing the type of `ctmd.getCompose` is but in any case the type appears to be incorrect. As is, it implies that `ctmd` was mutated after mapping over it with `concat(', rock on, Chicago')` – the type should remain unchanged.

I created a [jsbin](http://jsbin.com/yebusofivo/1/edit?html,js,console) to confirm.

```javascript
ctmd.getCompose.fork(
  console.log.bind(console, 'Error'),
  console.log
);
// Maybe {__value: "Rock over London"}
```

### DIFF
--- a/ch8.md
+++ b/ch8.md
@@ -736,7 +736,7 @@ map(concat(', rock on, Chicago'), ctmd);
 // Compose(Task(Maybe('Rock over London, rock on, Chicago')))
 
 ctmd.getCompose;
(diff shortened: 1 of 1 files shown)


DrBoolean:
Thanks!

PR #307, opened 2016-09-22
Add toArray function

samhavens:
Line 42 of `part1_exercises/support.js` calls toArray, which is not defined, so I wrote it. It uses the ES6 static `from` method on `Array`, which is supported in Node >= 4.

Oddly, this was causing me trouble when running tests in part 2.

### DIFF
--- a/code/part1_exercises/support.js
+++ b/code/part1_exercises/support.js
@@ -1,8 +1,9 @@
 //var curry = require('ramda').curry;
-//
-//
-//
-//
+
+function toArray(x) {
+  return x != null ? Array.from(x) : x;
+}
+
 function inspect(x) {
   return (typeof x === 'function') ? inspectFn(x) : inspectArgs(x);
 }


fix-fix:
See #297

samhavens:
I just saw that #297 solves this using ES5, but I feel this is more readable. And Node 4 is not an unreasonable criteria.

fix-fix:
This book is written for ES5.

In ES6 rewrite from #273 function `toArray` is no longer needed, because it was only used for defining `compose`, but exercises in ES6 version use `compose` from Ramda.

So I guess the ES6 version of `toArray` doesn't fit in any of those.

samhavens:
I see. Sorry, I should have noticed that everything else is ES5.

I think the book and exercises should be consistent about the use of
`compose`, `curry`, etc, or `_.compose`, `_.curry`, etc.

christiantakle:
@samhavens I'm pretty sure that the ES6 version of the book is consistent. If not please point them out and I will fix it.

Cheers

PR #309, opened 2016-09-26
ch9 update Array to use .concat rather than .push

exiadbq:
Explicit usage of Array.prototype.concat() API to enforce pure functions / no mutation in ch9.

### DIFF
--- a/ch9.md
+++ b/ch9.md
@@ -513,7 +513,7 @@ var ex3 = undefined;
 var addToMailingList = (function(list) {
   return function(email) {
     return new IO(function() {
-      list.push(email);
+      list.concat(email);
       return list;
     });
   };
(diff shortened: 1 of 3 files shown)


safareli:
This will not work `concat` returns new array and you are not using it's result.

Also as It's wrapped in  `IO` which is used to encapsulate side effects, so looks like it's intended to be impure.

exiadbq:
Oops, I have corrected the return value.  Thanks for that. 
Mutation can be avoided for the array operation to enforce the idea of pure function though.

DrBoolean:
I think the idea was that I was wrapping the push in IO to simulate a side-effect send. I like concat better anyways tho

exiadbq:
I like concat too, also concat can relate to semigroup and monoid... or relatives from afar.. lol.

Should it be merged?

safareli:
I think it's better to add comment before `list.push(email);` indicating that it's "simulating a side-effect" instead of making Pure, because then after some time, someone will ask why are we using IO When it's all pure :d

exiadbq:
Sounds good, let me close this one off.

Issue #264, opened 2016-04-01
Correct implementation of join for IO


DanBurton:
The provided implementation of `join` for `IO` is as follows:

``` javascript
IO.prototype.join = function() {
  return this.unsafePerformIO();
}
```

This is incorrect, because it runs the outer IO effect at "composition time" rather than "run time".

``` javascript
var example = new IO(function() {
  console.log('effect happened');
  return IO.of(3);
});

var joinedExample = example.join();
// prints 'effect happened'

joinedExample.unsafePerformIO();
// nothing is printed
joinedExample.unsafePerformIO();
// (again) nothing is printed
```

The correct implementation would be something like the following:

``` javascript
IO.prototype.join2 = function() {
  var thiz = this;
  return new IO(function() {
    return thiz.unsafePerformIO().unsafePerformIO();
  });
};

var joinedExample2 = example.join2();
// nothing is printed

joinedExample2.unsafePerformIO();
// prints 'effect happened'
joinedExample2.unsafePerformIO();
// (again) prints 'effect happened'
```

DrBoolean:
Good call! Want to PR or shall I?

DanBurton:
I went ahead and opened a PR. However, the `this` aliasing doesn't quite fit in with the style of the rest of the book, so perhaps you can come up with a cleaner implementation.

PR #317, opened 2016-12-13
Restructuring

christiantakle:
Change the shape of the original project to reduce duplication and make it easier to refine. 

### TODO
- [x] Check support url in FAQ.md
- [x] grep for `part1_`
- [x] grep for `part2_`

Consider fixing:
#264

#### Note
Markers do not work in the current version of 'include-codeblock'
https://github.com/azu/gitbook-plugin-include-codeblock/issues/33


DrBoolean:
Should we change the word "include" to "example 1.a" or something like that?

christiantakle:
@DrBoolean Not quite sure that I understand. Can you add an example on a specific line in the pr?

DrBoolean:
When we point to some code like here: https://github.com/MostlyAdequate/mostly-adequate-guide/pull/317/files#diff-8c7e026ecf14a664558c1415ad941841R346 the text says "include"

So at the end of this chapter: https://github.com/MostlyAdequate/mostly-adequate-guide/blob/restructuring/ch4.md it just reads as "include" rather than here is a link to the examples.

Or maybe i'm mistaking the concept. Totally a possibility

christiantakle:
Its the plugin I'm using it splices in the file that it points to.

Try
```sh
npm install
npm run setup
npm start
```

DrBoolean:
Ahh, gotcha, thanks!

DrBoolean:
Is there any way to rename include though for readers on github or does that need to be that way?

christiantakle:
Hmm not sure (https://github.com/azu/gitbook-plugin-include-codeblock#usage)

I was thinking that most would read it as the gitbook format at https://drboolean.gitbooks.io/mostly-adequate-guide/content/

DrBoolean:
Most definitely. If not, ebook is totally priority. Just curious since it would be a little less cryptic here.

KtorZ:
Any updates on that one?

christiantakle:
@KtorZ I havn't looked at it for a long time. I will see if there are any changes to the plugin I use or if there is an alternative.

If not I would be happy if we can break this down into something we can merge. Its getting hard to maintain.

christiantakle:
@KtorZ

I misunderstood what @DrBoolean meant. Its hard to make the include statement more readable one could hack the optional inlined key:value to provide some more information however I don't think its worth it.

Locally I did create a custom template in the JS-next branch where I added a link to the github source file. Not sure if its useful but that something you can decide weather you want or not.

screen shot 2017-08-14 at 22 54 55

Other news this https://plugins.gitbook.com/plugin/exercises seems more interesting to play with when we close this #317  and/or the #273

PR #318, opened 2016-12-13
`const` is a viable alternative to `Object.freeze`

SamuelMarks:
`const` wasn't introduced until ES6, whereas `Object.freeze` appeared in ES5:
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Statements/const
Ref: www.ecma-international.org/ecma-262/6.0/#sec-let-and-const-declarations

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze
Ref: http://www.ecma-international.org/ecma-262/5.1/#sec-15.2.3.9

So not sure if your guide has been updated to use features from newer standards…?

### DIFF
--- a/ch3.md
+++ b/ch3.md
@@ -58,13 +58,22 @@ In the impure portion, `checkAge` depends on the mutable variable `minimum` to d
 
 It might not seem like a lot in this example, but this reliance upon state is one of the largest contributors to system complexity(http://www.curtclifton.net/storage/papers/MoseleyMarks06a.pdf). This `checkAge` may return different results depending on factors external to input, which not only disqualifies it from being pure, but also puts our minds through the ringer each time we're reasoning about the software.
 
-Its pure form, on the other hand, is completely self sufficient. We can  also make `minimum` immutable, which preserves the purity as the state will never change. To do this, we must create an object to freeze.
+Its pure form, on the other hand, is completely self sufficient. We can  also make `minimum` immutable, which preserves the purity as the state will never change. To do this, we must create either create an object to freeze:
 
 ```js
 var immutableState = Object.freeze({
   minimum: 21,
 });
 ```
+
+Or use ES6's `const`:
+```js
+var checkAge = function(age) {
+  const minimum = 21;
+  return age >= minimum;
+};
+```
+
 
 ## Side effects may include...
 


ch3.md
```diff
@@ -58,14 +58,23 @@ In the impure portion, `checkAge` depends on the mutable variable `minimum` to d
...
 
 ```js
 var immutableState = Object.freeze({
   minimum: 21,
 });
 ```
 
+Or use ES6's `const`:
+```js
+var checkAge = function(age) {
```
SamuelMarks: `const checkAge` would work here instead of `var checkAge`; but not sure if that would confuse the reader.

fix-fix:
No, it's not. Example: https://jsbin.com/woxopuzeba/edit?js,console

What you have done is just replaced already unavailable internal closure variable, that can't be changed whether it's declared with `let` or `const` or `var`.

Thanks for your suggestion for using `const` though, it's indeed a good practice and is encouraged in ES6 version of the book, see #273

SamuelMarks:
Hmm yes of course that's right, silly me. Ah what put me off was the line:
> Its pure form, on the other hand, is completely self sufficient. We can also make `minimum` immutable, which preserves the purity as the state will never change.

Which could be solved using `Object.freeze` as you have shown, or alternatively [simply]:
```
const minimum = 21;
```

Wow #273 is a serious PR. I was thinking about going through all your chapters and replacing them, likely with a fancy `sed` script or an even fancier `estree`-utilising script [2to3 style]. But looks like someone's already done the hard yards :+1:

christiantakle:
@SamuelMarks If you have any comments/additions to #273 

Currently I want to update the shape of the original version of the book so its similar to my PR but still using ES5 for the Student facing code.

Then I will get back to my PR so I can finalise it and find a proper place for it to live.

I wonder if one could make it similar to a translation.

It would be nice with more interactivity so one could toggle the different version of Javascript. In the future one could cover other dialects.

SamuelMarks:
Yeah I'll take a look at some point. I'm still having fun experimenting with all kinds of different paradigms, everything from prototypical & class OO to functional (lazy, strict) to weird ones like Erlang & Pony.

I'm touching up on my C++, D, Go, Rust, Java, TypeScript, JavaScript and Python to try and figure out a decent paradigm to solve all my problems. Thus far I'm still stuck in the 'right tool for the right job' category, but will hopefully stumble upon a 'one size fits all' approach at some point.

In the meantime I'll be digging deep into compiler architecture to speed up multi-language multi-paradigm development through code-generation, and exploring the various optimisations done at various framework/language level approaches to problem solving. This is more than just conforming to design patterns and inlining thereof. But still seeking the answer!

[wow it's 2AM; I appear to of gone off on a rant >.<]

PR #331, opened 2017-06-08
Fix Left prototype map function

jontewks:
It looks like the map function wasn't filled out all the way, and so hopefully this is the implementation the author was intending to write, which it seems to be due to the way Left.of('rain').map is used on line 288.

Thanks!


jontewks:
Misunderstood it was intentionally like this for `Left`

PR #337, opened 2017-08-08
Remove offensive phrase from ch3.md

DanielPintilei:


snackycracky:
offensive phrase ? Could you pleas elaborate on that ?

Soreine:
Related #338 #339

DanielPintilei:
@snackycracky it's about "wrongful use of the name of the Lord your God", the third commandment from the Bible

snackycracky:
I think you are missinterpreting because actually only "god knows what", no?

ericpoe:
I vote for the neutral.

Religiously observant Jews would write "G-d" or "g-d" for similar reasons vowels are removed from the Tetragrammaton. Religiously observant Christians would see it as using the name of God in vain. Atheists and non-religious would see the usage of "god" as odd and would pull the reader away from the text. Non-native speakers might not be familiar with the idiom.

Avoid idioms when possible or go with neutral phrasing.

adicirstei:
Daniel,

You must be kidding. Have you read George Orwell's 1984?

snackycracky:
I am a true believer and I like the name of the lord three times in the book. It makes the book even more divine.

ferborva:
Believers or not, the fact is that this is a change that will leave the sentence in a state of general conformity. 
If the sentence hadn't included the word 'God' or 'god' in it from the beginning, no one would have said anything and the book would be     Just        As        Good.
+1 on my side to end this nonsense conversation that will end in god knows what.

snackycracky:
amen.

KtorZ:
I agree with @Soreine, @ericpoe and @ferborva comments. We can preserve the meaning (and the style) of the sentences while removing "offensive" content. 

Same goes for #338 & #339.

bitfrickler:
Isn't religion supposed to happen at home? As an atheist I am offended by you being offended.

KtorZ:
That's why we keep it out of the book ;)

PR #340, opened 2017-08-08
fix missing toArray

eliranmal:
replace a call to `toArray` (which declaration is missing), with a simple `Array.prototype.slice` invocation.

### DIFF
--- a/code/part1_exercises/support.js
+++ b/code/part1_exercises/support.js
@@ -39,7 +39,7 @@ function curry(fx) {
 }
 
 compose = function() {
-  var fns = toArray(arguments),
+  var fns = Array.prototype.slice.call(arguments, 0),
       arglen = fns.length;
 
   return function(){
--- a/code/part1_exercises/support.js
+++ b/code/part1_exercises/support.js
@@ -1,7 +1,7 @@
 //var curry = require('ramda').curry;
 
 
-var asArray = Function.call.bind(Array.prototype.slice);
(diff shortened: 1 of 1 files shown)


KtorZ:
What about adding the `toArray` definition instead since it's much more elegant :) ?
There's the `Array.from` method but I think it's ES6+ and we agreed to keep this version of the book compliant with ES5. 

Something like:

```js
const toArray = Function.apply.bind(Array.prototype.slice);
```

Will probably do, or anything you could come up with :)

eliranmal:
@KtorZ seems reasonable, i'll do that. and yes, i'll have it ES5 compliant, as i'm aware of the ES6 separation (and want to help with it!).
i'll just push another commit

eliranmal:
i chose to bind `call`, to have an API similar to slice (usages in the file rely on the ability to slice the array at various positions)

KtorZ:
Sounds fair :)
I am however going to be a bit picky but, what about calling it `toArray` like originally intended ^.^ ?

eliranmal:
oh whoops, that was an honest mistake (i have my own `asArray` in a whole other project..)

eliranmal:
@KtorZ all set :)

KtorZ:
@eliranmal Yup ;) .. I just want to test the code to make sure, for the sake of testing it before merging.

KtorZ:
Merged! :+1:

PR #343, opened 2017-08-23
show local mocha call from node_modules

pruett:
in the case that the user doesn't have `mocha` installed globally


KtorZ:
This is actually tricky since you can't really assume from where this should be ran. Users may also have their `NODE_PATH` pointing to the `.bin` directory too. 

I guess the spirit here is just to make it clear that `mocha` has to be used. Up to you to tweak a bit the line to match your own configuration :+1:

PR #351, opened 2017-11-01
Fix reversed_cars variable name

Hemaolle:
Looks like there was a mistake in the variable name since we're taking
the last car here, not reversing the list of cars.


KtorZ:
Indeed. Remains of old code I guess.

Thanks for the PR :+1:

PR #359, opened 2018-01-04
Simple typo fix: "amendable" -> "amenable"

ptrfrncsmrph:
"Amendable" might apply here(?), but I assume you were going for "amenable".


KtorZ:
I think you might be right on that one ^.^ ... 
I'd however rather wait for an extra opinion on the matter.

DrBoolean:
Sold!

PR #361, opened 2018-01-08
ES6 + Restructuring

KtorZ:
Finally!

This PR upgrade the book with ES6 / ES7 code and examples (based on the work of @christiantakle) and would close #273. I've opted for a style as close as possible to the airbnb styleguide as it seems to be the most accepted "standard" out there and give nice consistency to the JavaScript excerpt.

It also tackles #317 and will close a couple of opened issues about the exercises. As part of the restructuring:

- README has been slightly reviewed (mostly the title hiearchy)
- Firsts chapters are named with a leading 0 to increase consistency and ordering in many places.
- Exercises can now be ran directly within the browser (provided one reads through gitbook).
- Exercises can still be run via the CLI as an `npm` script (example npm run ch08)
- Exercises (and solutions) are now in separated files (exercise_a, exercise_b etc ..)
- Exercises are now grouped by chapters
- Exercises no longer rely on any dependencies (ramda or anything) to be more focus on teaching 
- There are 3 appendixes A, B, C providing implementations of algebraic data structure and functions used throughout the book or the exercises. 
- Link to ramda, folktale, fantasy-land or lodash are now given in annexes 
- The folder `code` has been renamed `exercises`
- Exercises are now listed in the SUMMARY
- Exercises provide more extensive feedback about errors (type mismatch, inversion of parameter in compose, mixing of IO.of() and new IO) based on issues and comments opened over last years

I think that's pretty much it.


DrBoolean:
🥇 !

Gonna take a bit, but i'll look this over for sure! Amazing!

DrBoolean:
My god. This is so great and I absolutely approve.

My only issue is with `Maybe`. We're using `Just` and `Nothing` even though they aren't really a thing here. I think I need to update it to be the real implementation and describe it as two subclasses of an abstract superclass. I avoided this at first because `Maybe` seemed to click with people where `Just/Nothing` didn't and it really gave the intuition without being too damaging.

I regret it, however, and would rather just bite the bullet and redo that section all together.

I'd like to merge and open an issue for that

KtorZ:
Ah! Good catch, It was somewhere in my head but I forgot to add the constructors.

I had in mind to leave the implementation as it is, but to provide a `nothing` and `just` helpers. We could also align the implementation with Either which would also make sense. Your call.

datenreisender:
I think there is an inconsistency in the code: When a function does not use the arguments handed to it, usually an empty argument list is used ([example](https://github.com/KtorZ/mostly-adequate-guide/blame/e945daef11f74990ca1da3c9ad24a43dabc01c74/ch08.md#L392)) but in some places an underscore is used ([example](https://github.com/KtorZ/mostly-adequate-guide/blame/e945daef11f74990ca1da3c9ad24a43dabc01c74/ch03.md#L192)). Is this done with a purpose I do not understand?

I suggest omitting the underscore variants `(_)`/`_` and use `()` when possible.

(I know, that strictly there is a diffenrence, because `(() => {}).length` is 0 and `(_ => {}).length` is 1, but this is rarely relevant and also seems irrelevant in the code here.)

KtorZ:
Hey!
In the first case, we actually want to create a closure and wrap the `x` inside an anonymous function. Thus, we aren't ignoring an argument per se, but simply creating a function with no argument (a pure side-effect). So the semantic is rather different than an omitted argument. 

The second case is actually a mistake (good catch :+1:) and should be empty a pair of empty brackets as well indeed. In some other cases though, the `_` usually refers to omitted arguments and act as a placeholder in the FP style. 

I'll fix this small oversight and invite you to open issues or PR  for any such case you may find / fix (such that we can have this PR merge and start iterating over it) :)

Thanks again!

datenreisender:
In Chapter 6 you moved the definition of `trace` but did not reflect this in the text. I suggest this little change: https://github.com/datenreisender/mostly-adequate-guide/commit/02c9ff0673f21c70cc0c918b6ec01a9be7dd6fe9

KtorZ:
Agreed, the phrasing makes more sense like this since the function has moved down a bit. 

Thanks for reviewing this :)

SUMMARY.md
```diff
@@ -1,58 +1,125 @@
...
+  * [Declarative Coding](ch06.md#declarative-coding)
+  * [A flickr of functional programming](ch06.md#a-flickr-of-functional-programming)
+  * [A Principled Refactor](ch06.md#a-principled-refactor)
+  * [In Summary](ch06.md#in-summary)
+* [Chapter 07: Hindley-Milner and Me](ch07.md)
+  * [What's your type?](ch07.md#whats-your-type)
+  * [Tales from the cryptic](ch07.md#tales-from-the-cryptic)
+  * [Narrowing the possibility](ch07.md#narrowing-the-possibility)
+  * [Free as in theorem](ch07.md#free-as-in-theorem)
+  * [In Summary](ch07.md#in-summary)
```
datenreisender: Above this line is a `  * [Constraints](ch07.md#constraints)` missing.

KtorZ: Thanks, fixed with 4be90f94d52e6411a35b35841edc4389b2fc4a0f

KtorZ:
@DrBoolean , any objection against me moving forward and merging this? I believe we can continue iterating on it but in the meantime, it will make this version available for new (and old) readers :+1:

DrBoolean:
Let's pull the trigger

PR #365, opened 2018-01-17
Bug: With sepia the links are not highlighted

datenreisender:
This is an old problem in GitBook as described in GitbookIO/gitbook#949.
I copied the workaround described there in https://github.com/GitbookIO/gitbook/issues/949#issuecomment-285365251

### DIFF
--- a/styles/website.css
+++ b/styles/website.css
@@ -1,0 +1,3 @@
+.book.color-theme-1 .book-body .page-wrapper .page-inner section.normal a {
+    color: initial;
+}


KtorZ:
Sounds reasonable; are the styles automatically picked up by Gitbook though (or do they need some extra configurations in the book.json) ?

datenreisender:
After `npm start` the style change was immediately active. So it seems that no extra configuration is needed.

PR #370, opened 2018-01-22
Update gitbook links

datenreisender:
If I understood it correctly, the gitbook now can be found at https://mostly-adequate.gitbooks.io/mostly-adequate-guide/.

The `in-summary` in one of the old links seemed to be a mistake since it did not lead to a correct page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,28 +23,28 @@ That said, typed functional languages will, without a doubt, be the best place t
 
 ## Read it Online
 
-For a best reading experience, [read it online via Gitbook](https://drboolean.gitbooks.io/mostly-adequate-guide/content/). 
+For a best reading experience, [read it online via Gitbook](https://mostly-adequate.gitbooks.io/mostly-adequate-guide/).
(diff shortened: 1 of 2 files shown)


KtorZ:
I wonder also if we shouldn't update all the links in the `SUMMARY.md` to also point to the hosted gitbook instead of the plain markdown files.

datenreisender:
I think keeping `SUMMARY.md` as it is is better so that the links also point at the files in the same location when looking at them locally or on GitHub.

Regarding how I found out: I saw the commit 228ea3d3a0c52e5e0a840d85f0e65444299b6160 and since the old Gitbook was not updated, I just tried the obvious name. 😄

benshine:
I noticed the old get book site is still up at https://drboolean.gitbooks.io/mostly-adequate-guide/ --would it be possible to take it down? Ideally also somehow direct readers to the correct version.
(meta-comment: what is the right place to make this suggestion?)

PR #374, opened 2018-01-27
Clarify curry example by matching the letter r and linking to regex d…

### LINKED ISSUE #372: Explanation of currying clearer if it used real chars instead of white space character class
```
match(/\s+/g, 'hello world');
// [ ' ' ]
```
In https://drboolean.gitbooks.io/mostly-adequate-guide/content/ch4.html
Makes perfect sense if the reader understands regular expressions and that \s+ means one or more spaces.  Consider rewriting with concrete characters. 
Sometimes js programmers don’t know regexes 😱

benshine:
Fixes  #372 , perhaps.

### DIFF
--- a/ch04.md
+++ b/ch04.md
@@ -31,13 +31,14 @@ const map = curry((f, xs) => xs.map(f));
 
 The pattern I've followed is a simple, but important one. I've strategically positioned the data we're operating on (String, Array) as the last argument. It will become clear as to why upon use.
 
-(The syntax just means match all of the in this string. Read [more about regular expressions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions) if you like.)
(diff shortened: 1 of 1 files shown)


KtorZ:
On the one hand I am glad to reduce the overall barrier, and in the other hand I am a bit sad that those things have to be explained this much ^^" 

I'll spend some time reviewing the new code and examples in the upcoming days. Thanks for the PR :)

benshine:
I erred towards little bit _too much_ explanation; the only confusing part of each example should be exactly what we’re trying to teach in that example. I almost used `’r’` Instead of the `/r/`  form because  that’s one less language construct to understand, but the _results_ of matching against the string are a little more complicated.

KtorZ:
And I totally agree with your statement :) 
You should give the reader the ability to focus 100% of his attention on the underlying topic and not on some extra quirks of JavaScript or anything else.

KtorZ:
Sorry for the time it took :+1:

PR #377, opened 2018-02-04
Corrections for "Pure Error Handling"

christianhg:
I was just rereading this section where I found some mistakes and inconsistencies.

By the way, I noticed a bunch of trailing whitespace in the file because my editor automatically fixes that. I wasn't sure if I could commit those fixes, but now you know :)


KtorZ:
Haha, thanks!
I'd rather fixed the last comment `  // 'Birth date could not be parsed' ` instead of all other occurrences but maybe that's just being lazy. 

About the white-spaces, please don't :). They're actually there to tell the markdown engine to go on the next line and not to collapse everything on the same line (yes, that's a mis-known and underrated feature of markdown). Maybe I'll switch to newlines at some point if this become a problem to enforce. Thanks for not committing it by the way!

PR #378, opened 2018-02-06
Monoids

DrBoolean:
Rough draft, no exercises yet

### DIFF
--- a/ch13.md
+++ b/ch13.md
@@ -1,0 +1,332 @@
+# Chapter 13: Monoids bring it all together
+
+## Wild combination
+
+In this chapter, we will examine *monoids* by way of *semigroup*. *Monoids* are the bubblegum in the hair of mathematical abstraction. They capture an idea that spans multiple disciplines, figuratively and literally bringing them all together. They are the ominous force that connects all that calculates. The oxygen in our code base, the ground on which it runs, quantum entanglement encoded.
+
+*Monoids* are about combination. But what is combination? It can mean so many things from accumulation to concatenation to multiplication to choice, composition, ordering, even evaluation! We'll see many examples here, but we'll only tip-toe on the foothills of monoid mountain. The instances are plentiful and applications vast. The aim of this chapter is to provide a good intuition so you can make some *monoids* of your own.
+
+## Abstracting addition
+
+Addition has some interesting qualities I'd like to discuss. Let's have a look at it through our abstraction goggles.
+
+For starters, it's a binary operation, that is, an operation which takes two values and returns a value, all within the same set.
+
+```js
+// a binary operation
+1 + 1 = 2
+```
+
+See? Two values in the domain, one value in the codomain, all the same set - numbers, as it were. Some might say numbers are "closed under addition", meaning the type won't ever change no matter which ones get tossed into the mix. That means we can chain the operation since the result is always another number:
+
+```js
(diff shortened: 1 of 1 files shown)


KtorZ:
Great! I'll look into that, but probably not before this weekend. Need any help / inspiration for the exercises?

KtorZ:
Haven't been though all of it yet. Still a pleasure to read. 

We'll have to update most example to be consistent with the new style, mostly missing semi colons and calls to `Either.of` instead of `Right`. I'll spend a bit of time carefully reviewing the code in the upcoming week :+1:

ch13.md
```diff
@@ -0,0 +1,332 @@
...
+    Endo(compose(run, other.run))
+})
+
+Endo.empty = () => Endo(identity)
+
+
+// in action
+
+// thingDownFlipAndReverse :: Endo [String] -> [String]
+const thingDownFlipAndReverse = fold(Endo(() => []), [Endo(reverse), Endo(sort), Endo(append('thing down')])
```
KtorZ: Shouldn't it be `fold(Endo(identity), [...])` here since we are folding on endomorphisms. Here we're just going to override the inputs given later to `[]` and apply every next functions to this empty array ^^"

ch13.md
```diff
@@ -0,0 +1,332 @@
...
+The default, intuitive behavior is to combine what a type is holding, however, there are cases where we ignore what's inside and combine the containers themselves. Consider a type like `Stream`:
+
+```js
+const submitStream = Stream.fromEvent('click', $('#submit'))
+const enterStream = filter(x => x.key === 'Enter', Stream.fromEvent('keydown', $('#myForm')))
+
+submitStream.concat(enterStream).map(submitForm) // Stream()
+```
+
+We can combine event streams by capturing events from both as one new stream. Alternatively, we could have combined them by insisting they hold a semigroup. In fact, there are many possible instances for each type. Consider `Task`, we can combine them by choosing the earlier or later of the two. We can always chose the first `Right` instead of short circuiting on `Left` which has the effect of ignoring errors. There is an interface called *Alternative* which implements some of these, well, alternative instances, typically focused on choice rather than cascading combination. It is worth looking into if you are in need of such functionality.
```
alessioscalici-olx: Typo: chose -> choose

ch13.md
```diff
@@ -0,0 +1,332 @@
...
+[1,2].concat([3,4]) // [1,2,3,4]
+
+"miracle grow".concat("n") // miracle grown"
+
+Map({day: 'night'}).concat(Map({white: 'nikes'})) // Map({day: 'night', white: 'nikes'})
+```
+
+If you stare at these long enough the pattern will pop out at you like a magic eye poster. It's everywhere. We're merging data structures, combining logic, building strings...it seems one can bludgeon almost any task into this combination based interface.
+
+I've use `Map` a few times now. Pardon me if you two weren't properly introduced. `Map` simply wraps `Object` so we can embellish it with some extra methods without altering the fabric of the universe.
```
ptrfrncsmrph:
```suggestion
I've used `Map` a few times now. Pardon me if you two weren't properly introduced. `Map` simply wraps `Object` so we can embellish it with some extra methods without altering the fabric of the universe.
```

kocikindra9 (commented):
ch13.md

PR #379, opened 2018-02-09
Fix summary

datenreisender:
`SUMMARY.md` contains a lot of minor errors: Sometimes entries were missing, sometimes the capitalisation differs, some links were wrong. 

After fixing some cases manually I got lazy, looked for a script, but the existing one generated summaries that differed too much, so I rolled a little Perl script on my own.

Invoke it with `npm run generate-summary`. Of course it would be nice to integrate it into a build process, but as far as I know, none exists. 

Oh, and in the first commit, I also fixed some more minor inconsistencies, e.g. some header levels were off in chapter 11.


KtorZ:
Yeay! Love the Perl script; we keep updating and fixing this summary so I am definitely for having it generated from the actual content! Great idea!

Few remarks on the PR however:

- In quite a few cases, you've updated the summary with by changing uppercases to lowercases. I'd rather do the opposite, and update the `ch*.md` to have the correct casing :+1: (would be amazing if you could do that).

- Secondly... please don't trim white-spaces on the exercises at the end of the chapters :'( ... Trailing white-spaces have an actual semantic in markdown and are seemingly used to force a newline. You're not the first doing that, hence I guess I'll move ahead and add a `.editorconfig` to avoid this kind of thing.

datenreisender:
Thanks for the comments, I will update the PR.

Did not know that trailing whitespace can be meaningful in markdown, but in that case I will update my own global `.editorconfig` which always trims them at the end to exclude Markdown: https://github.com/datenreisender/dotfiles/blob/ea336093c7dcf28d3286485e370a8440f4501d39/.editorconfig

datenreisender:
As you can see, I did fix the second aspect (wrong trimming of white-space). 

I am less sure about the capitalisation of the headings, as it currently is a little inconsistent. E.g. in chapter 6 “A flickr of functional programming” followed by “A Principled Refactor”. As described, e.g. in https://www.scribbr.com/academic-writing/capitalization-titles-headings/, there are of course different ways of capitalising headings. Would you like to have the same [capitalisation that is also traditionally found in titles](https://en.wikibooks.org/wiki/Basic_Book_Design/Capitalizing_Words_in_Titles) in the headings?

Currently, I only made the summary and the chapters consistent by always using the variant from the chapters. We could unify the chapters later and then correct the summary with the new script.

KtorZ:
Great work! 
About the capitalization, I'd rather stick with the Chicago Manual of Style or, The Elements of Styles that are, in my opinion, the two main reference on the English writing style. The [link](https://en.wikibooks.org/wiki/Basic_Book_Design/Capitalizing_Words_in_Titles) you found seems to actually summarize quite accurately the rules from the Chicago Manual; let's simply go over each title and apply them.

Though, you're right on that, this can be done in a separate PR / iteration :+1: (I'll open an issue about that).

generate-summary.pl
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env perl -n
```
KtorZ: What is the `-n` for in the shebang ? My system doesn't seem to like it that much

```
/usr/bin/env: ‘perl -n’: No such file or directory
```

removing it executes the script just fine.

KtorZ: I've move the call to perl in the npm script rather than executing the file as a standalone script. Works fine :+1:

datenreisender: Strange. The `-n` roughly tells Perl to run the script for every line of the input without automatically printing them. The switch is in Perl for ages, but it rather looks like your `env` works differently than mine. ¯\_(ツ)_/¯

PR #389, opened 2018-02-20
`inspect` function does not exist in `inspect method`

babakness:
Corrected references to missing `inspect` function from within `inspect` method.

### DIFF
--- a/ch08.md
+++ b/ch08.md
@@ -108,7 +108,7 @@ class Maybe {
   }
 
   inspect() {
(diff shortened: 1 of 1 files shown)


KtorZ:
Most (all) functions used in the book are detailed in the appendix. I think you are referring to:

https://github.com/MostlyAdequate/mostly-adequate-guide/blob/master/appendix_a.md#inspect

:)

PR #391, opened 2018-02-21
Minor erratum ch08.md

ptrfrncsmrph:
Removed the _inspect_ function call from inside the inspect method. Assuming this was a typo?


ptrfrncsmrph:
Saw this is a duplicate of most recent PR, my bad.

PR #392, opened 2018-02-21
Uppercase Milner in ch07.md title

ptrfrncsmrph:
Milner is proper noun (surname) and seems to have been mistakenly lowercased.


KtorZ:
Good catch :+1:

PR #393, opened 2018-02-21
Punctuation nitpicking ch04.md

ptrfrncsmrph:
Fix punctuation and stylize _higher order function_ definition to match definition of other terms.


KtorZ:
Nice :). Thanks for that.

PR #394, opened 2018-02-21
Type signature nitpicking ch08.md

ptrfrncsmrph:
I know we can play fast-and-loose with types in JS, but _technically_ `add` returns a number and `concat` expects a string (that is according to the docs in the appendix of this work and also in the Ramda docs, and the Ramda implementation of `concat` throws a `TypeError` when fed anything but a string). Ambiguous whether it should be explicitly noted `toString` is borrowed from Ramda, or if a simple version (`x => x.toString()`) should be added to appendix.

### DIFF
--- a/appendix_c.md
+++ b/appendix_c.md
@@ -182,8 +182,8 @@ const toLowerCase = s => s.toLowerCase();
 ## toString
 
 ```js
-// toString :: Number -> String
-const toString = n => n.toString();
+// toString :: a -> String
+const toString = x => x.toString();
 ```
 
 ## toUpperCase
--- a/appendix_c.md
+++ b/appendix_c.md
@@ -179,6 +179,13 @@ const take = curry((n, xs) => xs.slice(0, n));
 const toLowerCase = s => s.toLowerCase();
 ```
 
+## toString
+
+```js
+// toString :: Number -> String
+const toString = n => n.toString();
+```
+
 ## toUpperCase
 
 ```js
--- a/ch08.md
+++ b/ch08.md
@@ -312,7 +312,7 @@ Now, just like `Nothing`, we are short-circuiting our app when we return a `Left
 
 ```js
 // fortune :: Number -> String
-const fortune = compose(concat('If you survive, you will be '), add(1));
+const fortune = compose(concat('If you survive, you will be '), toString, add(1));
 
 // zoltar :: User -> Either(String, _)
 const zoltar = compose(map(console.log), map(fortune), getAge(moment()));


ptrfrncsmrph:
Absolutely, will do later today

ptrfrncsmrph:
Added a very simple toString to the appendix.

appendix_c.md
```diff
@@ -179,6 +179,13 @@ const take = curry((n, xs) => xs.slice(0, n));
 const toLowerCase = s => s.toLowerCase();
 ```
 
+## toString
+
+```js
+// toString :: Number -> String
+const toString = n => n.toString();
```
KtorZ: What about making it a bit polymorphic:

```hs
toString :: a -> String
```

ptrfrncsmrph: For sure! I'm still learning this stuff and before you brought that up, I had thought that a type signature like that was synonymous with `toString :: * -> String`, and would be wrongly implying a larger domain (since I'm really limited to types that have the toString method baked into their prototypes). But if I were to say `a` (as in, "there exists _some type_ that this works for") instead of `*` (as in, "I can put _any type_ in here and it'll work"), I'm in the clear (while still keeping the same function definition)?

KtorZ: Well, it would probably be safer to go for `String(n)` which will call `toString` for object where it exists but would also work for null types and other non "toString-friendly" types. 

In practice (i.e. in true FP languages like Haskell or PureScript that supports type-classes), you would define a class constraint on your type variable 'a'. This one would go as:

```hs
Show a => a -> String
```
Which reads as, given any type `a` that implements at least the interface `Show`, return a String. And that's how you get to constraint your inputs will keeping them polymorphic. We don't cover this in the book because it will be quite a mess (and counterintuitive) to implement type-classes in the system. For that, we need a proper super for static typing.

I'd say that, considering our scenario, `a -> String` combined with `n => String(n)` or simply `String` is good enough ^.^

ptrfrncsmrph: OK, I think I follow you, at least part of the way (but limited by my ignorance of Haskell/PureScript syntax). I did update the definition (in the commit after this one) to the more polymorphic:
```js
// toString :: a -> String
const toString = x => x.toString()
```
If it would be better written as
```js
// toString :: n -> String(n)
const toString = x => x.toString()
```
I can update to that (though I'm still a little unclear how to read that). Or if there is a different definition that would be preferable, let me know&mdash;I'm out of my element here (but thankful for the learning experience)!

PR #403, opened 2018-02-26
Add `reverse` function to appendix_c.md

ptrfrncsmrph:
`reverse` is referenced in some examples (such as towards end of Chapter 8), missing from appendix

### DIFF
--- a/appendix_c.md
+++ b/appendix_c.md
@@ -122,6 +122,16 @@ const reduce = curry((fn, zero, xs) => xs.reduce(fn, zero));
 const replace = curry((re, rpl, str) => str.replace(re, rpl));
 ```
 
+## reverse
+
+```js
+// reverse :: [a] -> [a], String -> String
+const reverse = x => 
+  Array.isArray(x)
+    ? x.reverse()
+    : x.split().reverse().join('')
+```
+
 ## safeHead
 
 ```js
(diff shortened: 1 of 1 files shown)


ptrfrncsmrph:
Yep, `reverse` is almost always used on arrays. There are two exceptions, one at end of Ch 8 (the diagram showing "hi" -> "ih") and beginning of Ch 10 when it is used on a string, but the type signature is given as `[a] -> [a]` elsewhere. String doesn't have a nice `String.prototype.reverse` method on it, that's why I was coercing it to an array with `split('')`, then calling `reverse`, then `join('')` to return a string. This definition would also work for strings (replacing `split` with the more explicit `Array.from`:
```js
s => Array.from(s).reverse().join('');
```
But I assume that's not what you meant. If we just want to support the type signature `reverse :: [a] -> [a]`, I think it suffices to say:
```js
xs => xs.reverse();
```
Right? Apologies if I'm being needlessly pedantic; I am still a student and this is how I learn 😀.

KtorZ:
> one at end of Ch 8 (the diagram showing "hi" -> "ih") and beginning of Ch 10 when it is used on a string,

Indeed. Thanks for digging that :) 
Then, it makes sense to be a bit polymorphic here and your implementation make sense. :+1:

KtorZ:
Merged with e3f5df02388ad2a5127071671b160c3becc15376

PR #404, opened 2018-02-26
Correct the order of arguments for `concat`

ptrfrncsmrph:
Corrected one of the last examples to match the `concat` examples earlier in the chapter, and as defined in appendix. Potentially confusing, I got hung up on this as it made it seem that `map` acts differently when operating on different functors.

### DIFF
--- a/ch08.md
+++ b/ch08.md
@@ -689,15 +689,15 @@ class Compose {
   }
 }
 
-const tmd = Task.of(Maybe.of('Rock over London'));
+const tmd = Task.of(Maybe.of(', rock on, Chicago'));
 
 const ctmd = Compose.of(tmd);
 
-map(concat(', rock on, Chicago'), ctmd);
-// Compose(Task(Just(', rock on, ChicagoRock over London')))
+map(concat('Rock over London'), ctmd);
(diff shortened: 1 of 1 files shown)


KtorZ:
You've a point ^.^ ... But what about doing it the other way around: i.e. updating the `Maybe.of` and `concat` calls such that they produce the example as shown ?

ptrfrncsmrph:
I came across a couple more examples where the order of arguments to `concat` is confused at beginning of Ch 9. I'll look for more, but maybe it would be more fitting to replace `concat` with `append` in these cases? And append an `append` to the appendix?
```js
// append :: String -> String -> String
const append = flip(concat)
```
(Since we already have `flip` and `concat`)

KtorZ:
I'll have to look a bit at what we did for the exercises. I recall there's one about using / creating this `append` function. Perhaps this steps can be skipped as the function would already be in the appendix.

ptrfrncsmrph:
Right, I believe you're thinking of Ch 5, exercise C. The `append` function isn't given outright, but the use of `flip` is suggested exactly towards that end. Maybe not worth giving it away by putting it in the appendix, but it also seems like half the time that `concat` is referenced throughout the book, `append` would be more fitting.

PR #407, opened 2018-02-28
Clarify order of evaluation

ceddlyburge:
The existing text states that "we are not encoding order of evalution", which does not make sense in the context of the code snippet above.

I have added some text to clarify the situation.

Another option would be to change the code snippet.


KtorZ:
Hmm. I believe that the paragraph doesn't really refer to the snippet above, but to declarative code in general. The clarification you added make it... clearer ^.^ 
Thanks for that, I am happily merging this.

ceddlyburge:
Thanks! I really enjoyed reading the guide by the way, it's excellently written ....

PR #415, opened 2018-03-07
More complex typos

ayeseeem:
Some slightly less simple typos I noticed while reading, working through the book.
I put them in a separate request in case you think they are debatable.


KtorZ:
Thanks :+1:

PR #416, opened 2018-03-07
A few tiny stylistic nitpics

ayeseeem:
A few small tweaks that I think improve the readability, but not major issues
I've kept these separate from the more uncontroversial simple typos :-)


KtorZ:
Sweet stuff :)

PR #417, opened 2018-03-07
Fix link to LICENSE - it's not .md

ayeseeem:
Alternatively I suppose we could rename the file LICENSE -> LICENSE.md, that would also make the link work


KtorZ:
I like the approach you had there. No point of making the LICENSE an *.md* file because.. it isn't.

PR #419, opened 2018-03-17
Fixed signature for map

HunderlineK:


KtorZ:
Hmmm... nope ^.^
You're describing the signature of `chain` here. `map` is indeed used to manipulate a value inside a functor within the same domain. 

What made you think that the signature was erroneous here?

HunderlineK:
@KtorZ All of the examples in ch8 that use map have a normal function as their argument. The first use of map:
`const streetName = compose(map(prop('street')), safeHead, prop('addresses'));`

Signature for prop:
`// prop :: String -> Object -> a`

`prop` itself is not `a -> f b` but simply `a -> b`

Granted, in ch8 we have

```
// map :: Functor f => (a -> b) -> f a -> f b
const map = curry((f, anyFunctor) => anyFunctor.map(f));
```

So I think the `map` in the appendix is not consisted with the map used in `ch8`

KtorZ:
Wow, my bad. I actually read the diff in the opposite direction >.> ....
You're right and the signature should indeed be `a -> b` !!

PR #422, opened 2018-03-21
Replacing functions with ones shown in example

rafaelmaza:
There are no functions named `hasSpaces` or `findSpaces` in the book or the exercises, so I think it was meant to be `hasLetterR` and `removeStringsWithoutRs` instead.

Repo search results for [hasSpaces](https://github.com/MostlyAdequate/mostly-adequate-guide/search?utf8=%E2%9C%93&q=hasSpaces&type=) and [findSpaces](https://github.com/MostlyAdequate/mostly-adequate-guide/search?utf8=%E2%9C%93&q=findSpaces&type=).

### DIFF
--- a/ch04.md
+++ b/ch04.md
@@ -58,7 +58,7 @@ been made available from the `exercises/support.js` module.
 
 ## More Than a Pun / Special Sauce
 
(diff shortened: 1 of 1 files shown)


KtorZ:
Good catch, this was updated recently with #374. This bit slipped through apparently :)

PR #423, opened 2018-03-24
Update ch03.md

demoran23:
Const ensures `minimum` is immutable, given its type.  Let allows reassignment.

### DIFF
--- a/ch03.md
+++ b/ch03.md
@@ -33,7 +33,7 @@ Let's look at another example.
 
 ```js
 // impure
-const minimum = 21;
+let minimum = 21;
 const checkAge = age => age >= minimum;
 
 // pure


KtorZ:
Hey, thanks for the inputs :) 
You're right that `let` allows reassignment and makes the variable mutable. Yet, that's the whole point in this small excerpt ^.^ 
This is done to emphasize on the fact that while out-of-scope, this variable is mutable and can potentially be changed (hence the non-purity of the function) as explain right below:

>  In the impure portion, `checkAge` depends on the mutable variable `minimum` to determine the result. In other words, it depends on system state which is disappointing because it increases the [cognitive load](http://en.wikipedia.org/wiki/Cognitive_load) by introducing an external environment.

Notice furthermore that using `const` doesn't make the variable *immutable* so to speak, but rather non reassignable. Using `const` on arrays or objects won't prevent them from changing :+1:

demoran23:
Perhaps I simply do not have the grasp of javascript that I think I do.  Could you explain to me how a const int is mutable.  It cannot be reassigned, and hence cannot be changed.  That's like the whole point of [const](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const).

KtorZ:
Unfortunately in JavaScript, assigning a variable isn't the only way to change the value of a variable. The most obvious example are arrays, even if declared with const, calling methods like `push` will mutate the array. For strings and number, if you start playing with the prototype, you can achieve quite crazy things. 
const simply prevent reassignation but in a dynamic language like JavaScript, it's a poor man defense ^^

PR #424, opened 2018-03-25
Call reduce on array in compose example

trivikr:
I tried to run the following code in browser console:
```javascript
```
 
It threw the following error:
```
Uncaught ReferenceError: reduce is not defined
    at <anonymous>:1:17
```

I think the method should be updated to
```javascript
const reverse = x => x.reduce((acc, x) => [x].concat(acc), []);
```
as `reduce` is a method on Array prototype ([MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce))

### DIFF
--- a/ch05.md
+++ b/ch05.md
@@ -32,7 +32,7 @@ Instead of inside to outside, we run right to left, which I suppose is a step in
 
 ```js
 const head = x => x[0];
-const reverse = reduce((acc, x) => [x].concat(acc), []);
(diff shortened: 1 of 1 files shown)


trivikr:
The method `reduce` seems to be defined in Appendix C at https://mostly-adequate.gitbooks.io/mostly-adequate-guide/appendix_c.html#reduce

This code change is not required, so closing.

PR #429, opened 2018-04-04
Fix unclear sentence.

harryvederci:


KtorZ:
I fail to see what was wrong with the previous sentence; And to me, this reformulation deviates from the original meaning, don't you agree ?

PR #430, opened 2018-04-04
Change const 'initials2' to 'initials'.

harryvederci:
Alternatively, the example could only call initials2 instead of initials (or both initials and initials2).


KtorZ:
Yup, I think that's an artifact remaining from the last big revision to ES6. :+1:
Thanks!

PR #432, opened 2018-04-07
Add an example of List.traverse with Either

ceddlyburge:
When I read chapter 12, I was unable to understand how `List.traverse` worked. 

I could have just believed it, but I wanted to understand, so I turned to the Appendix and read some other blog posts on the subject.

I feel that I have gained a lot from working it out and understanding how it is done. I have learned a little about the thought processes and style of coding in the functional world, and about the great power of Functor laws.

My intention with this addition is that I can help other people achieve these same benefits with a little less effort. I have tried to keep it as simple and concise as possible, and stick to the writing style of the book. I realise its quite a large addition though, so I won't be offended if you don't want to include it.

Cheers

Cedd

### DIFF
--- a/ch12.md
+++ b/ch12.md
@@ -118,7 +118,7 @@ This just runs a `reduce` on the list. The seed value is `of(new List([]))`, whi
 - `Right(a)`
 - `Left(a)`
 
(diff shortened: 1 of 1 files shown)


ceddlyburge:
Hi Matthias

I've used the variable names from the Annexe, and clarified the javascript notation.

Please let me know what you think.

Cheers

Cedd

PR #428, opened 2018-04-04
Fix typo.

harryvederci:
'By the by' -> 'By the way'.

PR #434, opened 2018-04-14
Fixed minor typo

BrutalCoding:
Obvious fix here, "By the by," should be "By the way,". After all, we are only human!


KtorZ:
No so obvious I believe ;) 
Duplicate of #428.

Thanks for reporting though!

PR #439, opened 2018-05-10
Ch 03 `const` -> `let`

cuyl:
`minimum` is a mutable variable

### DIFF
--- a/ch03.md
+++ b/ch03.md
@@ -33,7 +33,7 @@ Let's look at another example.
 
 ```js
 // impure
-const minimum = 21;
+let minimum = 21;
 const checkAge = age => age >= minimum;
 
 // pure


KtorZ:
I think this is actually correct and was wrongly changed to `const` while updating to ES6.
See the beginning of the next paragraph:

> In the impure portion, `checkAge` depends on the mutable variable `minimum` to determine the result.

The idea of that function is to emphasize on the fact that, having `minimum` defined outside of the function scope is potentially impure as minimum could change. Using `const` would prevent that to some extend but it'd be clearer to stick with `let` to leave all confusion out.

ptrfrncsmrph:
Aha, I stand corrected. I figured that the impurity arose from `checkAge` depending on a variable (mutable or not) defined outside of the scope of the function. This continues to be a point of confusion for me in regards to definition of purity.

ptrfrncsmrph:
Aha, I stand corrected. This continues to be a point of confusion for me in regards to definition of purity.

cuyl:
Also change next `minimum `?

```js
 const checkAge = (age) => {
   let minimum = 21;
  // maybe change minimum
   return age >= minimum;
 };
```

It still a pure function.

KtorZ:
The second `const` in the `checkAge` function is fine. We don't need to emphasize on the mutability here.

PR #443, opened 2018-05-15
Remove mutation and use reduceRight

AkshayIyer12:
Used ES6 syntax to reduce the code.


AkshayIyer12:
Inspect is removed as it was not referencing the inspect method of Maybe class. And when I tried using this.inspect it went into a recursive loop which exhausted the stack. The code is working right now with this commit.

PR #449, opened 2018-05-31
Small correction defining sequence

epeicher:
Not completely sure but I think the definition of `sequence` should be in terms of `traverse`

### DIFF
--- a/ch12.md
+++ b/ch12.md
@@ -53,7 +53,7 @@ See what has happened here? Our nested type gets turned inside out like a pair o
 
 ```js
 // sequence :: (Traversable t, Applicative f) => (a -> f a) -> t (f a) -> f (t a)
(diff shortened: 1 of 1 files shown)


KtorZ:
Curious to see your rational here ^.^
I think you're confusing a few stuff, looking at type signatures should already convince you. 

Note that here, we aren't defining `sequence` so to speak, but simply defining a pointfree version of the `*.sequence` that is available on some Applicative Functors. Having said that, I am 100% with you about using `traverse` when it comes to defining `sequence` (e.g. in the Identity Functor: https://github.com/MostlyAdequate/mostly-adequate-guide/blob/master/appendix_b.md#identity).

epeicher:
You are right, I was confusing a few stuff 😄 
Thanks for the explanation, it makes a whole lot of sense. I will remind to myself: use the type signatures Luke

PR #451, opened 2018-06-05
Fix typo (peak -> peek) in appendices A, B, and C

ryanlaws:
Intro text of summaries in all 3 appendices uses homophone "peak" instead of "peek" in "have a peak at..." - doesn't seem to be intentional to me, so it may be a typo.


KtorZ:
My bad :x ... Pardon my French!
Thanks for PR.

ryanlaws:
Absolutely - thanks for your work on the excellent book!

PR #456, opened 2018-06-26
Add cover image and title to generated EPUB

kottkrig:
Add a cover image (converted from `./images/cover.png`) and a proper title to the generated EPUB.


KtorZ:
Hey! Great stuff :tada: 
Is there any ways to specify the image location from the epub configuration? To be honest, I am pretty sure I remember a `cover.jpg` that we deleted / moved at some point. 
So, not relying on default location would make it clearer what this image is about and would avoid duplicating it in the meantime.

kottkrig:
The only documentation I can find regarding the cover image is [this section in the gitbook docs](https://github.com/GitbookIO/gitbook/blob/master/docs/ebook.md#cover):

> To provide a cover, place a cover.jpg file at the root directory of your book. Adding a cover_small.jpg will specify a smaller version of the cover. The cover should be a JPEG file.

A natural place for such a setting would be in the `book.json` file but there is no mention of such a property in the [docs](https://toolchain.gitbook.com/config.html) 😥.

KtorZ:
Pretty sure we can try `"cover": "images/cover.jpg"` in the config :p 

cf:

https://github.com/GitbookIO/gitbook/blob/b48f8472d35b27d549d7c9a5b61f77f6b78e1908/lib/output/ebook/getCoverPath.js#L14-L20

kottkrig:
Great find! That worked splendidly. 

It also seems to work with the original `.png` image. 👍

KtorZ:
Yeay :tada: !
Happily merging this (I fully trust you on the epub build though :upside_down_face:)

PR #457, opened 2018-06-27
Add npm scripts to generate epub and pdf

kottkrig:
The previous example in the readme is referencing a globally installed gitbook package which might not be installed on the user’s system. The new npm scripts uses the locally installed gitbook package in `node_modules`.

I also added the the generated pdf and epub file to gitignore.


KtorZ:
Great stuff, thanks!

PR #463, opened 2018-07-30
generate pdf's from markdown

johncmunson:
I often find that I learn foundational concepts better when I can print a copy and get away from the distractions of the computer. That's why I generated PDF's, to make it easier for folks like me.

The files were generated in Atom using the [markdown-pdf](https://atom.io/packages/markdown-pdf) plugin, which uses [mdpdf](https://github.com/bluehatbrit/mdpdf) under the hood.


KtorZ:
Hey @johncmunson !
Thanks for the PR although we already have generated PDF and ePUB for that purpose ^.^

(see https://github.com/MostlyAdequate/mostly-adequate-guide#download-it)

I may be missing something perhaps?

johncmunson:
Hah, what the heck I don't know how I missed that! Thanks!

Just curious, does that PDF get kept up to date and regenerated when there are updates to the `.md` files?

KtorZ:
Yes. The pdf is generated from the gitbook server which pulls changes as we merged them to master :+1:

So it should always be up-to-date.

PR #474, opened 2018-10-04
fix errorneous sample code of function composition

huangc28:
## Objective

 The sample code of the following:

```
const head = x => x[0];
const last = compose(head, reverse);

last(['jumpkick', 'roundhouse', 'uppercut']); // 'uppercut'
```

should be fixed with:

```
const head = x => x[0];
const reverse = arr => arr.reduce((acc, x) => [x].concat(acc), []);
const last = compose(head, reverse);

last(['jumpkick', 'roundhouse', 'uppercut']); // 'uppercut'

```

### DIFF
--- a/ch05.md
+++ b/ch05.md
@@ -39,7 +39,7 @@ Instead of inside to outside, we run right to left, which I suppose is a step in
 
 ```js
 const head = x => x[0];
-const reverse = reduce((acc, x) => [x].concat(acc), []);
+const reverse = arr => arr.reduce((acc, x) => [x].concat(acc), []);
 const last = compose(head, reverse);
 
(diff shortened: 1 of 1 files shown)


KtorZ:
Isn't that a nice example of currying :upside_down_face: ?

```js
// reduce :: (b -> a -> b) -> b -> [a] -> b
```

So, 

```js
// fn :: [a] -> a -> [a]
const fn = (acc, x) => [x].concat(acc);

// init :: [a]
const init = [];

// reduce' :: [a] -> [a]
const reduce' = reduce(fn, init)
```

So, everything looks fine!

```js
// reverse :: [a] -> [a]
const reverse = reduce'
```


No need for a fix :smile:

PR #475, opened 2018-10-04
Fix link to support.js

dwhelan:
Fix link to `support.js` in FAQ.md

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -104,7 +104,7 @@ A glossary/vade mecum will come at the end to synthesize all these notions.
 
 Most of the time, those methods are defined in specific vendor libraries such as `ramda` or
 `underscore`. You should also have a look at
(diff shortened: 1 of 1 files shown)


KtorZ:
Hi @dwhelan !
Thanks for catching this :tada:
However, now that we've introduced appendixes, I'd rather point to them instead of the current `support.js` which contains some plumbing necessary to have exercises running smoothly in browser and in node.js. That's quite some noise I don't want to bother readers with :open_mouth: 

Would you mind updating to point to the appendixes instead ?

VladimirLogachev:
I suppose that link to `support.js` was already fixed by #496 and this PR can be safely closed without merging.
@KtorZ, am I right?

KtorZ:
Indeed, fixed in #496

PR #478, opened 2018-10-06
typos - .ap() vs .map()

dhenson02:
could confuse newer JS dev readers


dhenson02:
lol good to know.  had only made it to chapter 4 which points us to Appendix A.  Thanks for the heads up

Issue #479, opened 2018-10-06
Nitpicking on example in chapter 8


eevman:
In "My First Functor", there is this example:
``` javascript
Container.of('bombs').map(concat(' away')).map(prop('length')); 
// Container(10)
```
Assuming that `concat` is defined as in appendix C:
``` javascript
const concat = curry((a, b) => a.concat(b));
```
`Container.of('bombs').map(concat(' away'))`
gives the result 
```
Container(' awaybombs')
```
and not
```
Container('bombs away')
```
which I suppose was the intended meaning...
Of course both give the final answer 10 :-)
Shouldn't the example be:
``` javascript
Container.of('bombs').map(append(' away')).map(prop('length')); 
// Container(10)
```
where `append` is:
``` javascript
const append = flip(concat);
```
or something similar?
Or is maybe `concat` defined as
``` javascript
const concat = curry((a, b) => b.concat(a));
```
but then it would be difficult to know what function `concat` does without looking at the implementation...
Any thoughts on this?

KtorZ:
Indeed, this was already reported by someone not so long ago. We should definitely have an `append` function here :)

eevman:
A followup question: shouldn't we add `append` to Appendix C? It is only mentioned in [example in Chapter 5](https://github.com/MostlyAdequate/mostly-adequate-guide/blob/f4a1c4bbcb2187047025af4665f3f5a3d8408d58/exercises/ch05/solution_c.js#L1)

I have found the [report](https://github.com/MostlyAdequate/mostly-adequate-guide/pull/404#issuecomment-369045206).

eevman:
Further on in chapter 8, there's this strange example:
``` javascript
const compLaw1 = compose(map(concat(' world')), map(concat(' cruel')));
const compLaw2 = map(compose(concat(' world'), concat(' cruel')));

compLaw1(Container.of('Goodbye')); // Container(' world cruelGoodbye')
compLaw2(Container.of('Goodbye')); // Container(' world cruelGoodbye')
```
This was changed in #191. I wonder why didn't they use `append` instead of changing the output to be correct in a strange way? :-)

PR #191, opened 2015-10-11
Update ch8.md: concat output incorrect

kapral18:
All the concat examples are incorrect, this one in particular.
Here's the plnkr http://plnkr.co/edit/GIdCCNwJT16J90B3jzP2 
or at least I am missing smth, am I?

PR #482, opened 2018-10-13
Ch08: Replace `concat` with `append` in one example.

eevman:
Fixes issue #479.

### DIFF
--- a/ch08.md
+++ b/ch08.md
@@ -67,7 +67,7 @@ Container.of(2).map(two => two + 2);
 Container.of('flamethrowers').map(s => s.toUpperCase()); 
 // Container('FLAMETHROWERS')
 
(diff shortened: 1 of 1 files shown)


KtorZ:
Hey, great stuff :) 
And I pretty much agree with everything you mentioned on #479. Are you interested in fixing those few other points as well? (Including #191 ?)

Incidentally, this will require to define this `append` in the appendix :+1:

eevman:
Sure, I'll do it. It's just that there might be additional cases of `concat` which should be corrected in chapters 10 and above which I have not yet read... :-)

KtorZ:
`append` has been added to the appendix as part of 45f716078748283f580e7348def467a7b27e17c5. So, moving on merging this.

Thanks again for the contribution :+1:

PR #486, opened 2018-10-29
Introduce the signature syntax "f x" in ch7

samuelpath:
See the following Twitter thread: https://twitter.com/drboolean/status/1056943736918167557

The signature syntax "f x" for "functor holding x" is used in chapter 8 without explanation.
We add it here and generalize it to other objects with which the reader is already familiar (like Promises or Arrays), so that when they encounter "f x" in ch8, it directly makes sense to them.

### DIFF
--- a/ch07.md
+++ b/ch07.md
@@ -99,11 +99,16 @@ const filter = curry((f, xs) => xs.filter(f));
 
 // reduce :: (b -> a -> b) -> b -> [a] -> b
 const reduce = curry((f, x, xs) => xs.reduce(f, x));
+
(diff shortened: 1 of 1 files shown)


DrBoolean:
Hi @samuelpath, this looks great. If possible, can we move it down to the constraints section? I think it's a great addition to lead into the functor stuff later and we should include it, but it belongs down with constraints since it is one.

samuelpath:
Hi @DrBoolean, I've moved the addition to the constraints section. 
I hope it's where you had it in mind.

PR #499, opened 2019-02-13
ch9: fix HTTP upload examples

eush77:
Hi,

I've been rereading this section of the book and found this bug I thought I killed 3 years ago in #253 alive and well.

The problem here is that the two uses of function `httpPost` do not match each other and the type signature regarding the "path" parameter. The fix is to either drop this parameter or add it to the type signature and the remaining use. This patch does the latter.

### DIFF
--- a/ch09.md
+++ b/ch09.md
@@ -295,7 +295,7 @@ Let's read a file, then upload it directly afterward:
 
 ```js
 // readFile :: Filename -> Either String (Task Error String)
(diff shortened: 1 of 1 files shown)


KtorZ:
Hmmm.. Indeed. This probably got lost in translation with the re-write to ES6 / modern-JS :thinking:

eush77:
I've just rebased the changes on master.

KtorZ:
Thanks!

PR #500, opened 2019-02-15
ch11: fix saveComment definition

eush77:
Hi,

Function `saveComment` in this chapter is defined as a composition of functions, one of which is `getValue('#comment')`. However, `getValue('#comment')` is not a function according to its type specification:

```js 
// getValue :: Selector -> Task Error (Maybe String) 
```

This patch fixes the definitions of `saveComment` to avoid type mismatch.

### DIFF
--- a/ch11.md
+++ b/ch11.md
@@ -27,7 +27,7 @@ Until now, we've managed to evade this common scenario with carefully crafted ex
 const saveComment = compose(
   map(map(map(postComment))),
   map(map(validate)),
-  getValue('#comment'),
+  _ => getValue('#comment'),
 );
 ```
 
@@ -169,7 +169,7 @@ const saveComment = compose(
   chain(eitherToTask),
   map(validate),
   chain(maybeToTask),
-  getValue('#comment'),
+  _ => getValue('#comment'),
 );
 ```
 


KtorZ:
Nice catch. What about using: https://github.com/MostlyAdequate/mostly-adequate-guide/blob/master/appendix_a.md#always ?

eush77:
I tried, but unfortunately we can't use `always` to wrap an object of type `a` into a no-argument function `() -> a`, since it waits for the second argument:

```js
always(1)    //=> function
always(1)()  //=> function
always(1)(2) //=> 1
```

So `always(getValue('#comment'))()` returns a function, whereas `(_ => getValue('#comment'))()` returns a Task.

KtorZ:
I am not sure to follow you on that one. We have:

```hs
always :: a -> b -> a
getValue :: Selector -> Task Error (Maybe String)
getValue("#comment") :: Task Error (Maybe String)
```

Hence:

```hs
always(getValue("#comment")) :: b -> Task Error (Maybe String)
```

Which should run perfectly fine in the pipeline; it's an unary function returning a `Task`.

eush77:
> it's an unary function returning a `Task`

Exactly, but as I recall from reading the book `()` in the type signature means that the function is *nullary,* meaning you call it like `saveComment()`, but in this case it would still return a function due to implementation details of `curry`:

```js
always(getValue("#comment"))() :: b -> Task Error (Maybe String)
```

I may be wrong about the `()` thing though, but I'm sure I've seen it used like that, either in the book examples or in the exercises.

KtorZ:
Hmmm. That is incorrect. The type for `always(getValue("#comment"))()` is actually `Task Error (Maybe String)`. Having `Unit` (a.k.a `()`) in a type signature just means that it only accepts one value which is precisely Unit. That's a type of cardinality 1 (there's only one possible value, which is just plain void). 

Functions don't usually take `()` as an argument, because it means that the argument is superfluous (there's nothing you can do with.. nothing). Still, a function written as `() -> a` still has one argument, which *has to be* Unit / void. In our case above, it's slightly more general than that because we don't have to require the argument to be unit; we ignore it anyway, so it could just be anything (hence the `b`).

eush77:
> The type for `always(getValue("#comment"))()` is actually `Task Error (Maybe String)`

It's not, and you can try it to see for yourself:

```js
> always(Task.of(1))
[Function: always]
> always(Task.of(1))()
[Function: always]
> always(Task.of(1))(undefined)
Task(?)
```

> Still, a function written as `() -> a` still has one argument, which _has to be_ Unit / void.

Translating to JavaScript, this means that `()` is a type consisting of a single value `undefined`, and all call sites must pass it explicitly. Perhaps there should be a paragraph somewhere in chapter 7 introducing this type before it's first used in chapter 8.

So the functions in this example must be called like `always(getValue("#comment"))(undefined)` and `saveComment(undefined)`, and thus `always` can indeed be used here to make the example point-free. I have just made this change in my branch.

I also found a couple of places where a function is called with no arguments, but I guess I will make another PR to fix that.

PR #502, opened 2019-02-18
Make nothing :: Maybe a, remove extra parameter

eush77:
`nothing` is only used in one place, and as a function it was called incorrectly before: `nothing()` instead of `nothing(undefined)`. (See https://github.com/MostlyAdequate/mostly-adequate-guide/pull/500#issuecomment-464747500.)

This PR fixes that by removing the extra layer of function wrapping entirely, and making `nothing` a constant of type `Maybe a`. This is to avoid the more verbose alternative — adding the `undefined` argument.

### DIFF
--- a/appendix_a.md
+++ b/appendix_a.md
@@ -141,8 +141,8 @@ const maybe = curry((v, f, m) => {
 ## nothing
 
 ```js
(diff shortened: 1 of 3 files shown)


KtorZ:
Good idea!

PR #503, opened 2019-02-18
ch09: fix signature and usage of logFilename in exercise_b

eush77:
This fixes type signature of function `logFilename` constructed in exercise_b, and changes its usage to match the type of the `()` parameter (see https://github.com/MostlyAdequate/mostly-adequate-guide/pull/500#issuecomment-464747500).


KtorZ:
Hey, good catch. I went for a slightly different approach and fixed the definition of `getFile` instead to be `IO ()`. I find it more a bit more intuitive than having these `() -> IO ()`

PR #504, opened 2019-02-19
ch12: fix exercises

eush77:
This fixes type signatures in the exercises, and also the definition of function `startGame`.


KtorZ:
Good catch! Thank you :tada:

PR #508, opened 2019-03-07
Fix ch07

VladimirLogachev:
### Fix reduce signature

Can't give `Array.proto.reduce` a curried function. The type states that it's an array, and the implementation uses its prototype, not some custom iterating function.

### Remove Promise.then type signature 

1. Promise is not usually used as an interface. Instead, most likely it should be described as a parametrized type *(without `=>`)*.

2. `Promise`'s `.then` method is not parametrically polymorphic.

```haskell
then :: (a -> b) -> Promise a -> Promise b
then :: (a -> Promise b) -> Promise a -> Promise b
```

You can't compile such function in **Haskell**, it leads to an attempt to create an infinite type 
``` haskell
type Promise a = a | Promise a
```
which is not possible.

So better completely remove this non-parametric example.

### DIFF
--- a/ch07.md
+++ b/ch07.md
@@ -97,13 +97,13 @@ const head = xs => xs[0];
 // filter :: (a -> Bool) -> [a] -> [a]
 const filter = curry((f, xs) => xs.filter(f));
 
-// reduce :: (b -> a -> b) -> b -> [a] -> b
+// reduce :: ((b, a) -> b) -> b -> [a] -> b
 const reduce = curry((f, x, xs) => xs.reduce(f, x));
 ```
 
 `reduce` is perhaps, the most expressive of all. It's a tricky one, however, so don't feel inadequate should you struggle with it. For the curious, I'll try to explain in English though working through the signature on your own is much more instructive.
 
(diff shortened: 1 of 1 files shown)


KtorZ:
Hello :) !
I am not sure to understand your remark regarding `reduce`? I see no problem with the use of a curried function with `reduce`, such as:

```js
const add = curry((a, b) => a +b);
reduce(add, 0, [1,2,3,4,5]);
// 15
```

The signature is maybe not 100% honest in the sense that the function doesn't **have to** be currified, and it would work with standard js functions too. So, both signature are valid for reduce though, the curried one fits more with regard to the rest of the book IMO.


About the `Promise` constraint, keep in mind that this is really an example. I agree that it's a bit confusing due to the behavior of `then` in JavaScript. I am not entirely in favor of removing the example, but reworking it maybe to something that is less confusing.

VladimirLogachev:
About reduce: If we give **truly** curried function, it should not work.

```js
const add = curry(a => b => a + b);
reduce(add, 0, [1,2,3,4,5]); // ?
```

VladimirLogachev:
Type is a contract. That means, "for any `a -> b -> a`".
There should be no exceptions.
Same thing with Promises.
Either we are telling how to use types the right way (and rely on them with no exceptions), 
or we are just same as ECMA people (who provide bad design).

DrBoolean:
I think this is good. Probably better to merge, be accurate, then re-add examples than to be inaccurate here.

VladimirLogachev:
@KtorZ what do you think about the PR?

KtorZ:
We, happy to merge, although I am a bit sad to see one example go away 😶
Yet, removing some source of confusion is a good idea.

PR #509, opened 2019-03-11
Ch09: fix example log function in section Mixing Metaphors

alaarmann:
Hello,

as a learner I got quite confused by `IO.of` instead of a constructor call in the definition of `log` in this example, because that way `log` doesn't get the correct type.
I do think it is just an erratum and should be changed to a constructor call as in later definitions of `setStyle` and `getItem`.


KtorZ:
Whoops! Good catch, thanks :+1:

Issue #510, opened 2019-03-12
Ch 8: safeHead() doesn't check for null input


thanhn2001:
```
// safeHead :: [a] -> Maybe(a)
const safeHead = xs => Maybe.of(xs[0]);
```

The claim is that `safeHead()` adds safety but it assumes that the input is an array and only checks if the first element is null. However, if the input `xs` itself is null then the attempt to reference the first element would result in an error. Would using map help?

```
const safeHead = xs => Maybe.of(xs).map(xs => xs[0]);
```

KtorZ:
Nope. That's the pitfall of using a dynamic type language like JavaScript. `safeHead` is safe regarding the `head` operation on a list. The type-signature, which acts here as an implicit contract, tells you that it expects a list. Give it something else, and you're violating the contract :boom: 

In practice, one could enforce some invariant and use a more defensive approach with something like:

```js
const safeHeader = xs => {
    assert(Array.isArray(xs), "not an Array");
    return Maybe.of(xs[0]);
};
```

Not passing an array is a programmer error and that should not be collapsed with the `Nothing` branch which is the result of a valid usage of that function. Enforcing invariant like this everywhere can be cumbersome and impractical however. Instead, if I'd recommend to use a type-checker or a statically typed language :)

KtorZ:
That was however a good question and it would be nice to add it to the FAQ!

PR #511, opened 2019-03-12
fix(ch08): make safe by wrapping input of safeHead with Maybe

thanhn2001:
Code change for #510.


thanhn2001:
Closing without merging. See explanation in #510.

PR #515, opened 2019-04-04
Fix pureLog in support.js to match the implemenation in the book

shaharke:
Specifically the implentation in support.js does not write the string to the console

### DIFF
--- a/exercises/support.js
+++ b/exercises/support.js
@@ -955,7 +955,7 @@ const validateUser = curry(function validateUser(validate, user) {
 
 const getFile = function getFile() { return IO.of('/home/mostly-adequate/ch09.md'); };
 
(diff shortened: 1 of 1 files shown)


shaharke:
Huh... I just realized that `pureLog` is implemented this way so that the exercises will pass...

KtorZ:
Hey! Indeed, the pure log from the support is a bit of a "fake" one; I'd rather fix the exercise wording to reflect that and keep the "testable" implementation here :+1:

shaharke:
Sure. You want me to close this PR and open a new one with a different wording? Happy to help

KtorZ:
So, the support implementation is indeed slightly different because we _need this_ in order to implement the validations in the exercises. Still, it makes sense to have the function actually logging something I suppose ^.^

PR #518, opened 2019-04-18
Improve exercises instructions

FernandoBasso:
From the README, it is not clear that the user should change into the
`exercises` directory and install the dependencies there. It is
important to mention this because there is indeed a `package.json` in
the `exercises` directory with a different set of dependencies and scripts to
assert the correctness of the exercises' solutions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,14 @@ All functions and algebraic data-structures presented in the book are gathered i
 $ npm i @mostly-adequate/support
 ```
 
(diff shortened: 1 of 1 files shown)


KtorZ:
Hey! Thanks for pointing this out, and sorry for not coming back earlier with this. I've actually added a small README in the `exercises/` folder to give clearer instructions there. PR are welcome to improve the README :smile: !

PR #519, opened 2019-05-02
CH05: Reverse needs an argument to be called upon

kbariotis:
`reverse` was not accepting any arguments and was throwing an error.

### DIFF
--- a/ch05.md
+++ b/ch05.md
@@ -39,7 +39,7 @@ Instead of inside to outside, we run right to left, which I suppose is a step in
 
 ```js
 const head = x => x[0];
-const reverse = reduce((acc, x) => [x].concat(acc), []);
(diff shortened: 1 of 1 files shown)


kbariotis:
Hmm I may have missed the point here. I see that that function must be referencing the one from https://mostly-adequate.gitbooks.io/mostly-adequate-guide/appendix_c.html.

KtorZ:
Indeed, functions in the book are defined in the appendixes (or our `npm` module [@mostly-adequate/support](https://www.npmjs.com/package/@mostly-adequate/support)

For example:
```js
> const { reverse } = require('@mostly-adequate/support');
> reverse("Mostly Adequate");
'etauqedA yltsoM'
```

PR #530, opened 2019-07-13
exercises/ch05/solution_c.js changed

bma3131:
exercises/ch05/exercise_c.js uses `concat` function. It also hints that the `flip` function may come in handy. But solution is done with `append` function and it doesn't use `flip`.


guilhermedeandrade:
I don't think this has to change, considering that  `append("foo")` and `flip(concat, "foo")` are the same

```js
// append :: String -> String -> String
const append = flip(concat);
```

KtorZ:
Thanks for catching this!
I've simply removed the hint in 3802257. 

Cheers!

PR #531, opened 2019-07-19
fix reverse function by adding an arg and arrow

guilhermedeandrade:
The reverse funciton @ line 42 was lacking an argument and the arrow or return call.

```js
                 ^
ReferenceError: reduce is not defined
````

So I just fixed by changing to the following:
```js
const reverse  = xs => xs.reduce((acc, x) => [x].concat(acc), []);
```

### DIFF
--- a/ch05.md
+++ b/ch05.md
@@ -39,7 +39,7 @@ Instead of inside to outside, we run right to left, which I suppose is a step in
 
 ```js
 const head = x => x[0];
-const reverse = reduce((acc, x) => [x].concat(acc), []);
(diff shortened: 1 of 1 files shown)


KtorZ:
Hey! This `reduce` function actually comes from the [Appendix C](https://github.com/MostlyAdequate/mostly-adequate-guide/blob/master/appendix_c.md#reduce) (as mentioned in the Chapter 04)! 

all code excerpt assumes that you have imported the appendixes (see also: https://www.npmjs.com/package/@mostly-adequate/support)

PR #533, opened 2019-07-30
fix: use concat instead of append

guilhermedeandrade:
At line 315, the `fortune` function was defined using `append`, which is not compatible with the expected behavior.
This PR suggests the usage of `concat` instead of `append`, so the output of the `fortune` function would be the expected one.


KtorZ:
Hey! Indeed! This was caught by someone else a bit earlier (#532) and now fixed! I believe I can still merge your contribution though!

PR #538, opened 2019-09-09
Ch08: fix findParam to make usage of Maybe reasonable

VladimirLogachev:
The text states:

> it's perfectly reasonable to have a `IO(Maybe([x]))`

But until this PR is merged, it seems not reasonable to have `Maybe` there, because there will be no `Nothing`, Only `Just []` (which does not make sense in my opinion). Also `findParam` has `find` in its name, which assumes that we expect **at most one** result from it.

Relevant changes in `findParam` function:
83ae24f392d1f5727268e199d3867fa0d902da47
acb7f35f542ad6f15ac5082d36945543ff5d8d64

Original type signature was `findParam :: String -> IO Maybe [String]`, which makes me think that `filter` was mistakingly left in code from the beginning.

### DIFF
--- a/ch08.md
+++ b/ch08.md
@@ -455,14 +455,14 @@ const toPairs = compose(map(split('=')), split('&'));
 // params :: String -> [[String]]
 const params = compose(toPairs, last, split('?'));
 
-// findParam :: String -> IO Maybe [[String]]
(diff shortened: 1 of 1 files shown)


DrBoolean:
I like it!

VladimirLogachev:
@KtorZ Could you please review and merge (if everything is fine)?

PR #540, opened 2019-10-04
Ch05: Updates reverse example to use newer array syntax

andyfleming:


VladimirLogachev:
Why is it important in this book?

andyfleming:
It's not that important. It's just a small update to keep up with the latest style of javascript. I personally think it reads a little better as well.

VladimirLogachev:
As far as I know, `...` means "get all the values from Iterator", and its usage seems to have more wide meaning, than concatenation.

PR #541, opened 2019-11-08
fix(ch05): hint change

braxtonchristensen:
- replace `flip` with `append`

The hint said, "Hint, the `flip` function may come in handy." After failing to make the exercise pass using flip a couple of times, I clicked to see the solution, it was using `append`. I am not sure if this hint was a red herring, but it confused me.


KtorZ:
Ah! This is indeed an artifact from the past!

Thank you :heart:

PR #552, opened 2020-01-16
update ch03.md

fedeghe:
Hi,
first, thx for this book... I love it
Then my cent: `slice` and `splice` have different purposes and functionality. I definitely agree that `splice` should be avoided due to its destructive nature. Anyway the only case when they (forgetting for a second about destruction side effect) have the same output is when the first parameter is exactly 0 and optionally both pass the same second integer, or, when only one parameter is passed. Not sure how to summarize that in a few words, but to avoid confusion I would at least shortly mention it.

### DIFF
--- a/ch03.md
+++ b/ch03.md
@@ -6,7 +6,7 @@ One thing we need to get straight is the idea of a pure function.
 
 >A pure function is a function that, given the same input, will always return the same output and does not have any observable side effect.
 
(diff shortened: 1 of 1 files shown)


fedeghe:
3 years guys for such a small thing, good luck
I dont even want that to appear in my history

PR #559, opened 2020-03-22
Fix firstWords composition

Croccifixio:
The `firstWords` composition (shown below) is incorrect.

```js
// firstWords :: String -> String
const firstWords = compose(join(' '), take(3), split(' '));
```

Instead of `join(' ')` it should be `intercalate(' ')`, which calls `Array.prototype.join()` as shown in its definition:

```js
// intercalate :: String -> [String] -> String
const intercalate = curry((str, xs) => xs.join(str));
```

### DIFF
--- a/ch12.md
+++ b/ch12.md
@@ -12,7 +12,7 @@ Let's get weird:
 // readFile :: FileName -> Task Error String
 
 // firstWords :: String -> String
-const firstWords = compose(join(' '), take(3), split(' '));
+const firstWords = compose(intercalate(' '), take(3), split(' '));
 
 // tldr :: FileName -> Task Error String
 const tldr = compose(map(firstWords), readFile);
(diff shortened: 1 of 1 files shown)


KtorZ:
Hey :) 
That definition is actually correct. We use `intercalate` from the appendix --> https://github.com/MostlyAdequate/mostly-adequate-guide/blob/master/appendix_c.md#intercalate and not `join` because `join` is already quite commonly used for "flattening monads". Hence, to avoid some confusion, this was renamed a while ago to `intercalate`. 

Cheers.

is-emmanuelo:
Hey,
I guess my wording is pretty horrible :). 
At the moment it's using `join(' ')`. This PR changes it to use `intercalate(' ')`

KtorZ:
Whola! My bad :sweat_smile: ... haven't got my morning coffee yet and I completely overlooked the actual diff. You - are - right. (thanks!)

PR #561, opened 2020-04-15
Update ch09.md

area73:
I think there was a missing map surrounding safeProp('street').

As it was stated on previous example on line 84


area73:
I'm colsing this issue because I was wrong 😅

PR #595, opened 2021-01-27
Adding missing find function, fixes #564

### LINKED ISSUE #564: Chapter 8 references 'find' function that is not present in the appendix
Chapter 8 has a section that references a `find` function:

    // findParam :: String -> IO Maybe [String]
    const findParam = key => map(compose(Maybe.of, find(compose(eq(key), head)), params), url);

But there does not appear to be a `find` function in the Appendix. I believe this may have referenced `filter` before (per issue #303) and there is a definition of `find` in there, so perhaps it just needs to be added to appendix?

Thanks for great book - enjoying very much so far!

dotnetCarpenter:
`find` is referenced in [chapter 8](https://github.com/MostlyAdequate/mostly-adequate-guide/blob/master/ch08.md#old-mcdonald-had-effects) and is needed for the `IO` example:

```js
// findParam :: String -> IO Maybe [String]
const findParam = key => map(compose(Maybe.of, find(compose(eq(key), head)), params), url);
```


dotnetCarpenter:
@KtorZ should I reinstate the white-spaces in _appendix_c.md_ before this PR can be accepted?

KtorZ:
@dotnetCarpenter no worries here. These are indeed extraneous. A double-whitespace marks a newline in markdown (probably should use an explicit <br> instead...) but in this case they are simply superfluous.

I've been busy with other stuff lately and didn't give much time to reviews which required more than a quick eye-check. I'll allocate some time soon-ish to go through some open PRs and issues on this project.

PR #596, opened 2021-01-31
change folktale link to data.task

dotnetCarpenter:
I did not fully understand the `Task` example in chapter 8 and following the link to Folktale 2 and the new `task` implementation made me even more confused.

I tried to change the example to work with [Folktale 2](https://folktale.origamitower.com/docs/v2.3.0/migrating/from-data.task/) but I do not know what to do with the error as it does not make sense in a `cleanup` operation.

Then I found [`data.task`](https://github.com/folktale/data.task) from the same author and with one minor change to the example everything worked as expected.

I added `String` because the `data` part of the `fs.readFile` callback is a `Buffer` in node version 14. _I am not sure if was a `string` at the time of writing the example._

**I think it is far easier to follow examples if they work with the current versions of contemporary software. Since `data.task` has been published as a separate module, it should be future proof with no apparent reason for change.**

PS: My editor automatically deletes dangling spaces, therefore there is a bunch of seemingly non-changed lines. But I think it is in everyone's interest, to not have dangling spaces littering the files.

### DIFF
--- a/ch08.md
+++ b/ch08.md
@@ -13,7 +13,7 @@ class Container {
   constructor(x) {
     this.$value = x;
   }
-  
+
   static of(x) {
     return new Container(x);
   }
@@ -61,13 +61,13 @@ Container.prototype.map = function (f) {
 Why, it's just like Array's famous `map`, except we have `Container a` instead of `[a]`. And it works essentially the same way:
 
 ```js
-Container.of(2).map(two => two + 2); 
+Container.of(2).map(two => two + 2);
 // Container(4)
 
-Container.of('flamethrowers').map(s => s.toUpperCase()); 
+Container.of('flamethrowers').map(s => s.toUpperCase());
 // Container('FLAMETHROWERS')
 
-Container.of('bombs').map(append(' away')).map(prop('length')); 
+Container.of('bombs').map(append(' away')).map(prop('length'));
 // Container(10)
 ```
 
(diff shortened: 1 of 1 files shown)


dotnetCarpenter:
In case it was not clear, my proposed change is:

<hr>

The internals are a bit too complicated to spill out all over the page here so we will use `Data.Task` (previously `Data.Future`) from Quildreen Motta's fantastic [data.task](https://github.com/folktale/data.task). Behold some example usage:

```js
// -- Node readFile example ------------------------------------------

const Task = require('data.task');
const fs = require('fs');

// readFile :: String -> Task Error String
const readFile = filename => new Task((reject, result) => {
  fs.readFile(filename, (err, data) => (err ? reject(err) : result(data)));
});

readFile('metamorphosis')
  .map(compose(split('\n'), String)) // Convert Buffer to String before split
  .map(head);
// Task('One morning, as Gregor Samsa was waking up from anxious dreams, he discovered that
// in bed he had been changed into a monstrous verminous bug.')

**Comment by sigfriedCub1990** [comment]
I can see that this is a PR from last year but that code is again obsolete. Just like @dotnetCarpenter I like to run things working, however this will not work with Folktale latest version, not to mention that the library is, unfortunately, discontinued.

In any case, for this to work we would need to modify it to:
```javascript
const { task } = require("folktale/concurrency/task");
const { split, head, compose } = require("@mostly-adequate/support");
const fs = require("fs");

const readFile = (fileName) =>
  task((resolver) => {
    fs.readFile(fileName, (err, data) =>
      err 
        ? resolver.reject(err) 
        : resolver.resolve(data)
    );
  });

  readFile("metamorphosis")
    .run()
    .future()
    .map(compose(split("\n"), String))
    .map(head)
    .map(console.log);
```
Hopefully this is helpful to anyone trying to run the example with the latest version of Folktale.

PR #599, opened 2021-02-07
Change `upload` param from String to Filename in ch 9

etmoore:
(nit)
iiuc, `readFile`'s first param is a `Filename`, so `upload`'s first param should also be a `Filename` to be consistent/correct.


KtorZ (approved):
Good catch :+1:

PR #603, opened 2021-02-20
fix(chapter4-exercise2): replace match function by test to fix filter…

whatthehanan:
the filterQs example is incorrect. currently, `match` method is being used which returns an empty array in case of no match and `filter` method considers it as a truthy value. I have replaced `match` with `test` for intended behavior

### DIFF
--- a/ch04.md
+++ b/ch04.md
@@ -148,7 +148,7 @@ Refactor to remove all arguments by partially applying the functions.
   
 {% initial src="./exercises/ch04/exercise_b.js#L3;" %}  
 ```js  
-const filterQs = xs => filter(x => match(/q/i, x), xs);
+const filterQs = xs => filter(x => test(/q/i, x), xs);
 ```  
   
 {% solution src="./exercises/ch04/solution_b.js" %}  
--- a/exercises/ch04/exercise_b.js
+++ b/exercises/ch04/exercise_b.js
@@ -1,4 +1,4 @@
 // Refactor to remove all arguments by partially applying the functions.
 
 // filterQs :: [String] -> [String]
-const filterQs = xs => filter(x => x.match(/q/i), xs);
+const filterQs = xs => filter(x => x.test(/q/i), xs);
--- a/exercises/ch04/solution_b.js
+++ b/exercises/ch04/solution_b.js
@@ -1,1 +1,1 @@
-const filterQs = filter(match(/q/i));
+const filterQs = filter(test(/q/i));
--- a/exercises/ch04/validation_b.js
+++ b/exercises/ch04/validation_b.js
@@ -1,7 +1,7 @@
 /* globals filterQs */
 
 filter.calledPartial = false;
-match.calledPartial = false;
+test.calledPartial = false;
 
 assert.arrayEqual(
   filterQs(['quick', 'camels', 'quarry', 'over', 'quails']),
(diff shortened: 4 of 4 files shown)


ch04.md
```diff
@@ -148,7 +148,7 @@ Refactor to remove all arguments by partially applying the functions.
   
 {% initial src="./exercises/ch04/exercise_b.js#L3;" %}  
 ```js  
-const filterQs = xs => filter(x => match(/q/i, x), xs);
+const filterQs = xs => filter(x => test(/q/i, x), xs);
```
dotnetCarpenter, line 151: There is no `test` function in _support/index.js_, so the above will fail without a definition.

dotnetCarpenter:
>match method is being used which returns an empty array in case of no match and filter method considers it as a truthy value. 

I don't think that is correct. `match` is defined as:
```js
// match :: RegExp -> String -> Boolean
const match = curry((re, str) => re.test(str));
```

That is not the [`string.match`](https://devdocs.io/javascript/global_objects/string/match) function you normally use in JavaScript, which returns an array. But a custom function that indeed returns a Boolean.

dotnetCarpenter:
Perhaps renaming the `match` function to `test` would make it easier to grok for js developers used to [`string.match`](https://devdocs.io/javascript/global_objects/string/match) and [`regExp.test`](https://devdocs.io/javascript/global_objects/regexp/test) .

exercises/ch04/exercise_b.js
```diff
@@ -1,4 +1,4 @@
 // Refactor to remove all arguments by partially applying the functions.
 
 // filterQs :: [String] -> [String]
-const filterQs = xs => filter(x => x.match(/q/i), xs);
```
dotnetCarpenter, line 4: This change seems appropriate. Again, renaming the library function `match` to `test`, as well as this change, would be less confusing for js devs.

ch04.md
whatthehanan: my bad. I hadn't seen the _support/index.js_ file and assumed the match definition was either the following as used little above the exercises section:

`var match = curry(function(what, str) {
  return str.match(what);
});`

or `_.match` from rambda that was missing an import. as implicated by: 

> We'll use a library called Ramda which curries every function by default.

I can close the PR or rename `match` to `test` as you suggested. I skimmed through Part 1 of the book and did not see any references to _support/index.js_ , I'd be willing to add those if they are indeed missing and not a mistake on my part. cheers!

exercises/ch04/exercise_b.js
```diff
@@ -1,4 +1,4 @@
...
+const filterQs = xs => filter(x => x.test(/q/i), xs);
```
dotnetCarpenter, line 4: This change seems appropriate. Again, renaming the library function match to test, as well as this change, would be less confusing for js devs.

ch04.md
dotnetCarpenter: Actually, when you run the solutions, you will use _exercises/support.js_ via `npm run ch04` etc, which has better console output.

I like the idea of renaming `match` to `test` and change the examples but that is just my opinion and I have little to say. I'm reading the book, just like you.

About the Ramda remark, it is a little curious since Ramda is never used in any of the examples in the book. Perhaps, it was a thought to include Ramda at some point but I really like implementations here because they are coded to show how the theory is applied and not performance, which seem to be a higher goal for Ramda (with good reason).

Issue #581, opened 2020-08-02
Ch12: array is not a monad


antonklimov:
`sequence(Either.of, [Either.of('wing')]); // Right(['wing'])` : how is this supposed to work. Array does not have `.sequence`.
`sequence(Either.of, new List([Either.of("wing")]))` might work here.

Later
```
const firstWords = compose(join(' '), take(3), split(' '));

// tldr :: FileName -> Task Error String
const tldr = compose(map(firstWords), readFile);

traverse(Task.of, tldr, ['file1', 'file2']);

```

Two problems here. First is again with array being treated as a monad. And second that `join` which is defined in support\index.js is a monadic join
```
// join :: Monad m => m (m a) -> m a
const join = (m) => m.join();
```
exports.join = join; 
whereas here the join needs to be an array join like:

```
// arrayJoin :: String -> [String] -> String
const arrayJoin = curry((sep, arr) => arr.join(sep));
exports.arrayJoin = arrayJoin;
```

KtorZ:
>  how is this supposed to work? Array does not have .sequence.

It is supposed to work because you only need the internal object to be an Applicative to write sequence / traverse. Here, the array ought to be treated as a Foldable and the `Either a` as the Applicative. Yet, I've just checked and indeed, the definition of the pointfree`sequence` is bonkers at the moment. 

> First is again with array being treated as a monad.

Nope, array is being treated as a Foldable.

> And second that join which is defined in support\index.js is a monadic join

Right! We have a name clash here, and this one should be [`intercalate`](https://github.com/MostlyAdequate/mostly-adequate-guide/blob/master/appendix_c.md#intercalate)

antonklimov:
It seems that `sequence` can be defined as:
`const sequence = curry((of, f) => traverse(of, identity, f));`
and `traverse` can be made aware of arrays like:
```
const traverse = curry((of, fn, f) =>
  Array.isArray(f)
    ? f.reduce((innerF, a) => fn(a).map(append).ap(innerF), of([]))
    : f.traverse(of, fn)
);
```
Would that work?

KtorZ:
Yes indeed! That sounds a bit more sound that extending the Array prototype with a "traverse" method ^.^

deleite:
[...]

PR #605, opened 2021-02-22
Fix sequence section in chapter 12

dotnetCarpenter:
This PR fixes most of the issues in #581.

From chapter 12, **Type Feng Shui** section, I have only changed line 4 from `sequence(Either.of, [Either.of('wing')]); // Right(['wing'])` to `sequence(List.of, Either.of(['wing'])), // List(Right('wing'))`, which unlike the original will not throw a TypeError. But also is an example of inverting a `Right (List (String))` to a `List (Right (String))`.
The original, `Right (List (String))` to `Right (List (String))` made little sense.

Line 5, however is not changed and does not highlight anything particular useful, since `sequence(Task.of, left('wing'))` = `Task.of(left('wing'))`.

```js
sequence(List.of, Maybe.of(['the facts'])); // [Just('the facts')]
sequence(Task.of, new Map({ a: Task.of(1), b: Task.of(2) })); // Task(Map({ a: 1, b: 2 }))
sequence(IO.of, Either.of(IO.of('buckle my shoe'))); // IO(Right('buckle my shoe'))
sequence(List.of, Either.of(['wing'])), // List(Right('wing'))
sequence(Task.of, left('wing')); // Task(Left('wing'))
```

Only two changes was required to _support/index.js_.

1. `Right`
```js
traverse(of, fn) {
  return fn(this.$value).map(Either.of); // added return
}
```
2. `Map`
```js
  // ----- Pointed Map
  static of(x) {
    return new Map(x);
  }
```

### DIFF
--- a/ch12.md
+++ b/ch12.md
@@ -45,7 +45,7 @@ Let's rearrange our types using `sequence`:
 sequence(List.of, Maybe.of(['the facts'])); // [Just('the facts')]
 sequence(Task.of, new Map({ a: Task.of(1), b: Task.of(2) })); // Task(Map({ a: 1, b: 2 }))
 sequence(IO.of, Either.of(IO.of('buckle my shoe'))); // IO(Right('buckle my shoe'))
-sequence(Either.of, [Either.of('wing')]); // Right(['wing'])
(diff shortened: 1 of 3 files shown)


dotnetCarpenter:
Blocks #607

PR #606, opened 2021-02-23
fixing Waltz of the Types first example

dotnetCarpenter:
By using the improved `traverse` function from https://github.com/MostlyAdequate/mostly-adequate-guide/issues/581#issuecomment-668274810
and adding `toString` since the nodejs `fs.readFile` returns a [`Buffer`](https://devdocs.io/node/buffer#buffer_class_buffer) and not a `String`.

```js
// readFile :: FileName -> Task Error String

// firstWords :: String -> String
const firstWords = compose(intercalate(' '), take(3), split(' '), toString);

// tldr :: FileName -> Task Error String
const tldr = compose(map(firstWords), readFile);

traverse(Task.of, tldr, ['file1', 'file2']);
// Task(['hail the monarchy', 'smash the patriarchy']);
```

Before the `traverse` function threw a `TypeError: f.traverse is not a function`, on `
traverse(Task.of, tldr, ['file1', 'file2']);`.

Thanks to @antonklimov.

### DIFF
--- a/ch12.md
+++ b/ch12.md
@@ -12,7 +12,7 @@ Let's get weird:
 // readFile :: FileName -> Task Error String
 
 // firstWords :: String -> String
-const firstWords = compose(intercalate(' '), take(3), split(' '));
+const firstWords = compose(intercalate(' '), take(3), split(' '), toString);
 
 // tldr :: FileName -> Task Error String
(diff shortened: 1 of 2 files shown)


dotnetCarpenter:
Blocks #607

PR #644, opened 2023-12-05
Ch06: Fix link to script source

timciep:
Fix link to script source, which was not working on the GitBook site, by using the absolute GitHub URL.

### DIFF
--- a/ch06.md
+++ b/ch06.md
@@ -167,7 +167,7 @@ And we're done!
 <img src="images/cats_ss.png" alt="cats grid" />
 
 Here is the finished script:
-[include](./exercises/ch06/main.js)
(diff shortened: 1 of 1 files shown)


KtorZ (approved):

ferllop:
First of all excuse me if this not the proper place (or way) to communicate this..

This change modifying the include path of an exercise on chapter 6 now breaks the include when generating pdf and epub files and also when running gitbook on my local. The previous version was working ok for me in pdf, epub and gitbook.

Also note that in the public gitbook version of the book, the link is pointing to "mostly-adequate" repository instead of "mostlyadequate" (note the dash). I don't know how public gitbook is managed and from where came that "extra dash" but this dash seems to be the real error, not that the relative path was not working. 

Maybe the solution is to revert this pull request and to update the public version with a corrected relative link and additionally show chapter 13 as well.

Greetings!

PR #645, opened 2023-12-07
Update build workflow to run properly

ferllop:
# Changes
- the old ubuntu version was causing the build workflow to stay queued forever. Updated to 22.04
- use a variable to store the book filename on build workflow
- use updated syntax on build workflow
- run generate-summary on build worflow


ferllop:
One of the latest pull request modifying the include path on chapter 6 now breaks the include when generating pdf and epub files and also when running gitbook on my local. The previous version was working ok for me.

Note that on the public gitbook version of the book the link is pointing to "mostly-adequate" repository instead of "mostlyadequate" (note the dash).
Maybe the solution is to update the public version, as it is not showing chapter 13 as well.

I don't wanted to include that correction on this pull request as the previous wanted to correct some sort of error.

KtorZ (approved):
Thanks.

PR #649, opened 2024-08-14
Chapter 6

spencerolson:


exercises/ch04/exercise_a.js
```diff
@@ -1,4 +1,9 @@
 // Refactor to remove all arguments by partially applying the function.
 
 // words :: String -> [String]
-const words = str => split(' ', str);
+// Original code
+// const words = str => split(' ', str);
+//
+
+// const words = str => split(' ', str);
+const words = split(' ');
```
KtorZ, line 9: Pointfree style is only mentioned in Chapter 5, so the code is purposely written in a non-pointfree style here. The change is therefore ill-advised.

exercises/ch04/lessons_learned.md
```diff
@@ -0,0 +1,24 @@
...
+
+  ```javascript
+  b.someFunction(a)
+  ```
+
+  that's equivalent to calling the curried function
+
+  ```javascript
+  someFunction(a)
+  ```
```
KtorZ, line 24: Not really. This is only true if `someFunction` doesn't rely on `this`. Also has little do to with currying I am afraid, so this is off-topic for Chapter 4.

spencerolson:
Accidental pull request, apologies. Thanks for the feedback, though :)

